PR #2, opened 2018-01-24
Timeslider

lindsaycarr:
~The one issue I am having is putting the slider at the correct location on the map. I ended up hard coding a value in because `chart_height - (chart_height * 0.15)` to bring it 15% from the bottom edge did not work. It seems that the map has decided to have different dimensions than what I setup for the SVG. Hoping to discuss this at the meeting later today.~

The slider is discrete (snaps to the years provided). At this point, nothing happens when the slider is changed, but I think we could easily return `round_yr` from the `dragged` function to subset data based on the slider actions.

slider

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -1,6 +1,6 @@
 // Width and height
 var chart_width     =   1000;
-var chart_height    =   600;
+var chart_height    =   800;
 
 // Projection
 var projection = d3.geoAlbers()
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -22,6 +22,9 @@ var map = svg.append("g")
 d3.queue()
   .defer(d3.json, "data/conus_map.geojson")
   .await(create_map);
+  
+// dummy var for now
+var years = [1950, 1960, 1970, 1980, 1990, 1995, 2000, 2005, 2010, 2015];
 
 function create_map() {
   
@@ -33,5 +36,6 @@ function create_map() {
 	var state_data = arguments[1];
   
   add_states(map, state_data);
+  add_timeslider(map, years, chart_width, chart_width);
   
 }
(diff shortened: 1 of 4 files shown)


js/build_map.js
```diff
@@ -33,5 +36,6 @@ function create_map() {
 	var state_data = arguments[1];
   
   add_states(map, state_data);
+  add_timeslider(map, years, chart_width, chart_width);
```
lindsaycarr: wait ... I think I just solved my slider placement issue. I was passing chart_width in twice so it was using the wrong value for chart_height inside this function. Fix on its way!

lindsaycarr:
Chart_height issue fixed. Now located here:
image

js/timeslider.js
```diff
@@ -18,7 +18,7 @@ function add_timeslider(map, years, chart_width, chart_height) {
   
   var slider = map.append("g")
     .attr("id", "slider")
-    .attr("transform", "translate(0," + 575 + ")");
+    .attr("transform", "translate(0," + (chart_height - (chart_height * 0.05)) + ")");
```
aappling-usgs: this could be `chart_height * 0.95` instead, if desired

js/build_map.js
```diff
@@ -36,6 +36,6 @@ function create_map() {
 	var state_data = arguments[1];
   
   add_states(map, state_data);
-  add_timeslider(map, years, chart_width, chart_width);
+  add_timeslider(map, years, chart_width, chart_height);
```
aappling-usgs: ah hah!

js/timeslider.js
lindsaycarr: oh duh! I will fix in a future PR

PR #3, opened 2018-01-25
State zoom

aappling-usgs:
Similar state formatting to previous water use viz, gentle state highlighting on hover, zoom in and out of individual states, URL changes with view and can be used to recover zoom to specific state.

state-zoom


jread-usgs:
😂

PR #4, opened 2018-02-09
Timeslider changes title

lindsaycarr:
Added a title to show timeslider changing data in some way + indicated where other functions would go in order to change the actual data shown.

### DIFF
--- a/js/timeslider.js
+++ b/js/timeslider.js
@@ -39,7 +39,16 @@ function add_timeslider(map, years, chart_width, chart_height) {
               dragged(dragging_circle, xscale, min_year, max_year);
               dragging_circle.classed("dragging", true); 
             })
-            .on("end", function() { return d3.select(this).classed("dragging", false); })
+            .on("end", function() { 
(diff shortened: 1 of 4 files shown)


js/timeslider.js
```diff
@@ -39,7 +39,16 @@ function add_timeslider(map, years, chart_width, chart_height) {
...
             })
-            .on("end", function() { return d3.select(this).classed("dragging", false); })
+            .on("end", function() { 
+              var dragged_circle = d3.select(this);
+              var new_year = xscale.invert(dragged_circle.attr("cx"));
+              
+              // this is where functions to update data would go
+              // just pass in `new_year` which is the year the 
+              // timeslider has been changed to
+              updateTitle(new_year);
```
lindsaycarr: @aappling-usgs this is where an `updateCountyBoundaries` function might go. You would need to know if people were zoomed into the state or not though.

aappling-usgs: great! yep, we'll know that.

PR #5, opened 2018-02-12
Historical county boundaries

aappling-usgs:
It's not beautiful yet. There's more to do with data processing:
0) Get the county boundaries dataset onto S3 so everybody can play. Should we do this with the raw file straight from IPUMS NHGIS, or should we do it with a partly processed version so not everybody has to wait for the full spatial data processing workflow?
1) Integrate the bash script, scripts/process/topo_county_boundaries.sh, into the automatic vizmake build
2) Somehow still need more cleanup to make polygon directions uniform? Even though I think I've corrected this both with `sf::st_read(check_ring_dir=TRUE)` and with `geo2topo` (i thought it did automatic clockwising unless `--no-force-clockwise`?), the shape transitions don't make it look like everything is clockwise...
3) Build counties datasets for more than just AL and AZ
4) The county boundaries are now a little simpler than the state boundaries. maybe we should use the merged county boundaries for the state boundary when we zoom in? also possible/likely that we should futz with the degree of simplification more; this occurs in scripts/process/topo_county_boundaries.sh

And there's more to do with prettifying: as noted in (2) above, transitions don't look like they're always going clockwise to clockwise. or maybe it's something about the transform (zoom) being applied at a different time or over a different transition period than the county bounds drawing?

However, there are county boundaries that I can draw for AZ and AL, and they change when the slider-selected year changes.

county-zoom-180211


aappling-usgs:
Hey @lindsaycarr , as noted above, this isn't beautiful, and it's not even super shareable, but could you review so that i'm not coding alone here? maybe just comment and don't merge, because this PR requires you to have some big data locally and requires a partially manual workflow (bash script and some file copying are not integrated, as noted above).

Also, i may not be able to get much more done before friday's meeting, so we should think about how and whether to try to hand this off to you or @mwernimont for a few days. I know Marty is pretty busy, not sure about you...and maybe this is enough for friday, anyway...what do you think?

js/counties.js
```diff
@@ -0,0 +1,73 @@
...
+    
+    // download the data and run the processing function
+    oneStateCounties = d3.json("data/" + stateFIPS + "-quantized.json", function(error, oneStateCounties) {
+      if(error) callback(error);
+      countyData.set(state, oneStateCounties);
+      callback(error, countyData.get(state));
+    });
+  } else {
+    // if we already have the data, just run the processing function
+    callback(null, countyData.get(state));
```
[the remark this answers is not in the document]
aappling-usgs: hmm. could do that...i used `callback` because it's a fairly common argument name (e.g., is the name of the 2nd arg to `d3.json`, https://github.com/d3/d3-request#json) and makes sense to me because it's the function that gets called after the data loading happens, just as in d3.json. so `callback` feels descriptive to me. but i'm open to something else - what would be clearer to you?

scripts/fetch/state_boundaries.R
```diff
@@ -22,5 +22,14 @@ fetch.map_data <- function(viz){
   map_args <- append(viz$fetch_args, viewbox_args)
   map_data <- do.call(get_map_data, args = map_args)
   
+  # revise the ID column to be compatible with county data and concise hashes
+  browser()
```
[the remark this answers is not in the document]
aappling-usgs: whoops, nope, thanks!

js/counties.js
```diff
@@ -0,0 +1,73 @@
+function hideCounties() {
+  map.selectAll('.county')
+    .data([])
```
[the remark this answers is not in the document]
aappling-usgs: well, it seems to work, and made sense to me because if the data array is empty then you'd figure everything would be in the .exit() selection, right? but i suppose an alternative would be to simply `map.selectAll('.county').remove()` - yep, just tried it and it works fine. i'll switch to that.

aappling-usgs:
based on our convo at stand-up, i went ahead and took the `transition()` line out of the county boundaries updating statement in this commit: https://github.com/USGS-VIZLAB/water-use-15/pull/5/commits/543be44e522d4ff52a55c153909f8c0f47271c44

js/counties.js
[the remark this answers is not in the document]
lindsaycarr: Well, this might just be a convention I need to learn! Keep it because it sounds like changing it might not be beneficial to anyone since it's a standard practice.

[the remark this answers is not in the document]
aappling-usgs: OK, keeping it for now. Happy to discuss again later if it continues to not work for you. Even though `callback` isn't a keyword, the whole `callback` idea is pretty specific to javascript because it's tied to the asynchronous execution thing. Over time I hope we'll continue to become more comfortable with it...and maybe also find better ways to name our callback arguments in the process =).

scripts/process/clean_county_boundaries.R
```diff
@@ -0,0 +1,255 @@
...
+    list(files=files, year=year)
+  })
+  names(map_shps) <- sapply(map_shps, function(map_shp) map_shp$year)
+  
+  # load into R
+  decade_shps <- lapply(map_shps, function(shapefile_files) {
+    message("reading county data for year ", shapefile_files$year)
+    shp_file <- grep("\\.shp$", shapefile_files$files, value=TRUE)
+    layer <- sf::st_layers(shp_file)$name
+    one_year_sf <- sf::st_read(shp_file, layer=layer, stringsAsFactors=FALSE, quiet=TRUE, check_ring_dir=TRUE)
```
lindsaycarr: I'm running this right now and I'm getting an error on this line for the very first file (1950):
```
one_year_sf <- sf::st_read(shp_file, layer=layer, stringsAsFactors=FALSE, quiet=TRUE, check_ring_dir=TRUE)
Error in st_sf(x, ..., agr = agr) : 
  no simple features geometry column present
```
Updated my `sf` package from 0.5 something to 0.6-0 and that went away (also updated `sp` to 1.2-7, but don't remember what the prev number was)

scripts/process/clean_county_boundaries.R
```diff
@@ -0,0 +1,255 @@
...
+                                   paste('County', county_FIPS))),
+      county_long = ifelse(!is.na(county_long), county_long,
+                           ifelse(!is.na(county_short), county_short,
+                                  paste('County', county_FIPS)))) %>%
+    select(-FIPS_U)
+  return(counties)
+}
+
+consolidate_county_polygons <- function(state_shp) {
+  # check shape validity and fix if needed/possible
```
dblodgett-usgs: Note I've gotten some mileage out of this "st_make_valid" from this package: https://github.com/r-spatial/lwgeom

scripts/process/clean_county_boundaries.R
```diff
@@ -0,0 +1,255 @@
...
+  # save to one big zip file
+  oldwd <- setwd(geojsondir)
+  on.exit(setwd(oldwd))
+  zipfile <- file.path(oldwd, viz[['location']])
+  if(file.exists(zipfile)) file.remove(zipfile)
+  filestozip <- dir()
+  zip(zipfile, files=filestozip)
+}
+
+simplify_combine_shps <- function(decade_shps) {
```
dblodgett-usgs: I'm confused about the use of the word simplify here. This is just simplifying attributes right?

scripts/process/topo_county_boundaries.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
dblodgett-usgs: (this is mostly Dave being curious...) Not following why we need a bash script to do this. -- oh!! upon a little googling, I see. The toposimplify implements the "Visvalingam’s" algorithm. Looks like that algorithm is available in the mapshaper package... https://github.com/ateucher/rmapshaper -- but I guess that topoquantize function is important too?

scripts/process/clean_county_boundaries.R
aappling-usgs: yes, simplifying attributes here.

scripts/process/clean_county_boundaries.R
aappling-usgs: thanks for the link. i'm hesitant to try that one because it appears to be tough to install on windows, and if we ever end up needing new collaborators to run this code, the extra install headache doesn't seem to be worth it yet. they say `st_buffer(x,0)` works pretty well most of the time, which so far seems to be the case here. https://github.com/r-spatial/sf/issues/487. but yeah, if that ends up being insufficient, good to keep `st_make_valid` in mind.

scripts/process/topo_county_boundaries.sh
aappling-usgs: Yes, and topo2geo and toposimplify and topoquantize are all said to be faster at the command line than via javascript (node.js), so i've been assuming that's also the case for R. also, having being converted to the D3 world, i don't see a big reason to prefer an R solution over Mike Bostock's widely used solution.

scripts/process/clean_county_boundaries.R
dblodgett-usgs: Yeah. The buffer usually fixes self intersection errors, but it only goes so far. It wasn't terrible to install on mac, but I believe it would be hard on windows...

scripts/process/topo_county_boundaries.sh
dblodgett-usgs: 👍

PR #6, opened 2018-02-13
custom projection in place

mwernimont:
So basically just has the projection call the function from the custom_projection.js file instead of geoAlbers(), and the function does the rest.  Maybe wanna start testing it by adding Alaska, Hawaii, and Puerto Rico, but for now its positive it's working.  Also, the Custom Projection JS file always needs to be loaded before build_map.js or it won't find the function to project the map.

### DIFF
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -15,8 +15,7 @@ function albersUsaPr() {
       .rotate([157, 0])
       .center([-3, 19.9])
       .parallels([8, 18]);
-
-  // XXX? You should check that this is a standard PR projection!
+      
   var puertoRico = d3.geoConicEqualArea()
       .rotate([66, 0])
       .center([0, 18])
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -1,0 +1,147 @@
+// A modified d3.geo.albersUsa to include Puerto Rico.
+function albersUsaPr() {
+  var ε = 1e-6;
+
+  var lower48 = d3.geoAlbers();
+
+  // EPSG:3338
+  var alaska = d3.geoConicEqualArea()
+      .rotate([154, 0])
+      .center([-2, 58.5])
+      .parallels([55, 65]);
+
+  // ESRI:102007
+  var hawaii = d3.geoConicEqualArea()
+      .rotate([157, 0])
+      .center([-3, 19.9])
+      .parallels([8, 18]);
+
+  // XXX? You should check that this is a standard PR projection!
+  var puertoRico = d3.geoConicEqualArea()
+      .rotate([66, 0])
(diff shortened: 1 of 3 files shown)


js/custom_projection.js
```diff
@@ -0,0 +1,147 @@
...
+      .center([-2, 58.5])
+      .parallels([55, 65]);
+
+  // ESRI:102007
+  var hawaii = d3.geoConicEqualArea()
+      .rotate([157, 0])
+      .center([-3, 19.9])
+      .parallels([8, 18]);
+
+  // XXX? You should check that this is a standard PR projection!
```
lindsaycarr: who's "you" here? Is that something that still needs to be done?

js/custom_projection.js
```diff
@@ -0,0 +1,147 @@
+// A modified d3.geo.albersUsa to include Puerto Rico.
+function albersUsaPr() {
```
lindsaycarr: this function name seems misleading since it is adding alaska and hawaii, right?

viz.yaml
```diff
@@ -77,18 +77,23 @@ publish:
...
+                  "map_css", "hashes", "projection", "map_utils", "timeslider"]
       sections: ["map"] # adding header and footer here, doubled each on the page
   -
     id: url_hashes
     location: "js/url_hashes.js"
     mimetype: application/javascript
+  -
+    id: custom_projection
+    location: "js/custom_projection.js"
+    mimetype: application/javascript
```
lindsaycarr: Can we add a comment right here about how this needs to come before `build_map`? Then, it's recorded in the actual code where someone might change it, and not just in your PR description.

js/custom_projection.js
mwernimont: That was the author's comment, It projected PR just fine, so I don't think its anything to worry about.

js/custom_projection.js
mwernimont: I guess I don't see the misleading as Alaska and Hawaii are apart of the USA, what would you rather have it?  A simple change is always easy.

js/custom_projection.js
lindsaycarr: can you remove the comment?

js/custom_projection.js
ldecicco-USGS: "a part" vs "apart" ...it's only now dawned on my that the space can make a pretty significant difference :smile_cat:

lindsaycarr: I'm looking at this a second time .... it is less misleading now that I read it again. So nevermind :)

js/custom_projection.js
mwernimont: Sure

PR #7, opened 2018-02-21
add mock 2015 wu data to s3 + viz.yaml

lindsaycarr:
Only used select up to this point for "cleaning". Not sure about best format for matching up with county data yet.

### DIFF
--- a/scripts/s3_bucket_setup.R
+++ b/scripts/s3_bucket_setup.R
@@ -3,7 +3,7 @@
 
 library(aws.signature)
 message('check that credentials for dev profile at ', aws.signature::default_credentials_file(), " match those in get_dssecret('dev-owi-s3-access')")
-aws.signature::use_credentials(profile='dev', file=aws.signature::default_credentials_file())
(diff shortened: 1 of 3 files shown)


scripts/s3_bucket_setup.R
```diff
@@ -12,3 +12,8 @@ put_bucket(new_bucket_name, region='us-west-2', acl='private') # gives error if
 
 # this command posted the data (took 1.5 hrs)
 put_object(file='data/nhgis0002_shape.zip', object='IPUMS_NHGIS_counties.zip', bucket='viz-water-use-15')
+
+# this command posts the USGS WU mock up data from Google Drive (see pinned items in water-use-fy18-viz Slack channel)
+# had to delete the first row of info before uploading (version num + authors)
```
aappling-usgs: interesting! do you think that was a constraint on the AWS side or on the aws.s3 side?

aappling-usgs: lindsay clarifies: " it uploaded just fine without doing that. I just wanted to easily use the built in excel reader within vizlab and not have to add a whole extra reader just to skip line 1"

PR #8, opened 2018-02-22
Proto spatial

aappling-usgs:
image

OK! Achieved:
1. acquired county boundaries for everywhere including PR and VI
2. computed centroids for all counties
3. created centroid points for arizona counties in map

However:
4. had to use a simpler projection and/or just Arizona counties (not sure which was needed in the end) to get the projection not to error out on something about a `stream`. 

So next steps (for someone else - @mwernimont ?) are:
5. try again to use the custom projection for the county points (i basically just gave up)
6. progressively add back in states until we find the limits of the custom projection

i think the problem may be something to do with some counties having extreme coordinates that weren't expected by the projection, but i'm really not sure. in the meantime, enjoy the purple circles!

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -33,6 +33,7 @@ d3.queue()
   .defer(d3.json, "data/state_boundaries.geojson")
   .defer(d3.json, "data/states.json")
   .defer(d3.json, "data/counties.json") // could load this later
+  .defer(d3.json, "data/county_centroids.json")
   .await(create_map);
 
 // dummy var for now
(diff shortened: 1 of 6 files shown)


aappling-usgs:
requesting reviews from both of you because @mwernimont is the custom projection expert and @lindsaycarr is likely to want to use this PR ASAP

lindsaycarr:
I'm going to merge this because I really want that centroid calculation stuff, and the projection issues might take longer than we want for this prototype. I'll probably go ahead and just try to make this in R as a quicker step. Comments on the projection issues should still go here but I've recorded that as issue #9 so that we don't forget.

PR #10, opened 2018-02-22
Scaling centroids into static maps

lindsaycarr:
Creates 5 difference PNGs and puts them into `cache/`. See examples on Slack.

### DIFF
--- a/scripts/process/compute_centroids.R
+++ b/scripts/process/compute_centroids.R
@@ -23,13 +23,14 @@ process.compute_centroids <- function(viz) {
   centroid_spdf <- sp::SpatialPointsDataFrame(centroids, spdat@data, match.ID=FALSE)
   
   # limit even more because centroids are a problem for our d3 projection?
-  centroid_spdf <- centroid_spdf[centroid_spdf$STATEFP %in% '04',]
+  # centroid_spdf <- centroid_spdf[centroid_spdf$STATEFP %in% '04',]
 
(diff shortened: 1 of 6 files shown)


scripts/process/compute_centroids.R
```diff
@@ -23,13 +23,14 @@ process.compute_centroids <- function(viz) {
   centroid_spdf <- sp::SpatialPointsDataFrame(centroids, spdat@data, match.ID=FALSE)
   
   # limit even more because centroids are a problem for our d3 projection?
-  centroid_spdf <- centroid_spdf[centroid_spdf$STATEFP %in% '04',]
+  # centroid_spdf <- centroid_spdf[centroid_spdf$STATEFP %in% '04',]
```
lindsaycarr: I didn't want to completely delete this so that we could still remember it when we continue to iterate on the implementation working w/ d3 and the projection

scripts/process/compute_centroids.R
```diff
@@ -23,13 +23,14 @@ process.compute_centroids <- function(viz) {
...
 
   # filter out Northern Mariana Island (STATEFP=='69'), Guam (STATEFP=='66'), and American Samoa (STATEFP=='60')
   centroid_spdf2 <- centroid_spdf[!(centroid_spdf$STATEFP %in% c('69','66','60')),]
   # limit to the necessary columns
   centroid_spdf3 <- centroid_spdf2[, c('GEOID')]
   
-  # write to file
-  geojsonio::topojson_write(centroid_spdf3, file=viz$location)
```
lindsaycarr: put this into a secondary step, see `process/sp_to_topojson.R`

Issue #35, opened 2018-02-27
Create tooltip function


aappling-usgs:
Enable #17 through D3 by attaching tooltip functions to each county polygon and/or dot.

aappling-usgs:
read json from #17

lindsaycarr:
2 hrs so far

lindsaycarr:
1 more hour as part of review/fix/discuss process.

Issue #17, opened 2018-02-26
Basic tooltip content


aappling-usgs:
County name and water use value for now. Follows on #35

aappling-usgs:
need a json file that contains county name and water use value and FIPS for mapping to the county polygons

aappling-usgs:
Total 2 hours

Issue #13, opened 2018-02-26
Get new data onto S3 (replace mock stuff)


aappling-usgs:

lindsaycarr:
Took < 1 hr

Issue #9, opened 2018-02-22
Fix projection weirdness with centroids


lindsaycarr:
See #8

lindsaycarr:
@mwernimont how long has this taken you so far?

mwernimont:
Shoot, I would say 4hrs? I'll do better with other task on timekeeping.

aappling-usgs:
Remaining: just need to get Alaska working and check against state polygons when those are available.

mwernimont:
Alaska STATEFIP: 02 COUNTYFIP: 016 has coordinates landing it in Canada which causes the projection cut off the dot which throws an error in D3 and stops all dots from being projected.  A quick fix at the moment is filtering out that COUNTYFIP and correcting it when we have time.  PR here https://github.com/USGS-VIZLAB/water-use-15/pull/69

aappling-usgs:
resolved by #75

PR #39, opened 2018-02-27
Getting water use data to a json file

ldecicco-USGS:
Working on #17

I'm not entirely sure this is what's needed (probably at this state has more than what is actually needed). But....it might be (also depends a bit on #35 and #13)

### DIFF
--- a/scripts/process/json_wu_data.R
+++ b/scripts/process/json_wu_data.R
@@ -1,0 +1,7 @@
+process.json_wu_data <- function(viz) {
+  deps <- readDepends(viz)
+  wu_data_15_simple <- deps[["wu_data_15_simple"]]
+  
+  # write to file
+  write(rjson::toJSON(wu_data_15_simple), file=viz[["location"]])
+}
--- a/scripts/process/json_wu_data.R
+++ b/scripts/process/json_wu_data.R
@@ -1,7 +1,7 @@
 process.json_wu_data <- function(viz) {
   deps <- readDepends(viz)
   wu_data_15_simple <- deps[["wu_data_15_simple"]]
-  browser()
+
   # write to file
   state_names <- unique(wu_data_15_simple$STATE)
   wu_list_for_json <- lapply(state_names, function(s) {
--- a/scripts/process/simplify_wu_data.R
+++ b/scripts/process/simplify_wu_data.R
@@ -7,10 +7,24 @@ process.simplify_wu_data <- function(viz) {
   orig_wu_type_col <- deps[["col_names"]][["orig_names"]]
   new_wu_type_col <- deps[["col_names"]][["new_names"]]
   
-  columns_to_keep <- c("STATE", "STATEFIPS", "COUNTY", "COUNTYFIPS", "FIPS", "YEAR")
+  columns_to_keep <- c("STATE", "STATEFIPS", "COUNTY", "COUNTYFIPS", "FIPS", "YEAR") # also double as character cols
   columns_to_rename <- c("TP-TotPop", orig_wu_type_col)
   new_names <- c("countypop", new_wu_type_col)
   
+  # none of the columns we kept had characters (-- or N/A) that caused the entire column to
+  # be read in as character and needed to be converted back to numeric, but in case
+  # we add more data as we go along, here is the code that does that:
+  which_numeric_cols <- !names(wu_df) %in% columns_to_keep
+  numeric_data <- wu_df[, which_numeric_cols]
+  fixed_numeric_data <- as.data.frame(sapply(numeric_data, function(v) { 
+    has_missing_symbol <- grep("--|N/A", v)
+    v[has_missing_symbol] <- NA
+    v <- as.numeric(v)
+    return(v)
+  }))
+  wu_df[, which_numeric_cols] <- fixed_numeric_data
+  
+  # now select only the pertinent columns
   wu_df_sel <- wu_df[, c(columns_to_keep, columns_to_rename)]
   names(wu_df_sel) <- c(columns_to_keep, new_names)
   
--- a/viz.yaml
+++ b/viz.yaml
@@ -38,7 +38,9 @@ info:
     sf:
       repo: CRAN
       version: 0.6.0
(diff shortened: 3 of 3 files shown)


ldecicco-USGS:
Thus far, I spent 1 hour on this task (less...but we're rounding)

ldecicco-USGS:
Should mention...I just realized the purple circles are gone...did this PR do that?

aappling-usgs:
It wasn't your PR. I think it was Lindsay's PR that I merged last night, though she didn't create the problem. Probably it's just that Lindsay added back in the other states worth of data other than Arizona, because what's happening is the browser is trying and failing to create the circles, failing on projecting the ~7th point in the dataset. Until we resolve the projection issue (#9 ) this will probably continue to happen, but I'm OK with that because we have the issue created and assigned already.

ldecicco-USGS:
PHEW!

viz.yaml
```diff
@@ -38,7 +38,9 @@ info:
     sf:
       repo: CRAN
       version: 0.6.0
-      
+    rjson:
```
lindsaycarr: I've never heard of `rjson`! I've only used `jsonlite`.

viz.yaml
```diff
@@ -254,6 +264,7 @@ publish:
       counties_dictionary: "counties_dictionary"
       county_boundaries: "county_boundaries"
       county_centroids: "county_centroids_topojson"
+      wu_data_15_json: "wu_data_15_json"
```
lindsaycarr: and this for sure makes it end up in `target/data`, correct?

scripts/process/json_wu_data.R
```diff
@@ -0,0 +1,7 @@
+process.json_wu_data <- function(viz) {
+  deps <- readDepends(viz)
+  wu_data_15_simple <- deps[["wu_data_15_simple"]]
+  
+  # write to file
+  write(rjson::toJSON(wu_data_15_simple), file=viz[["location"]])
```
lindsaycarr: I think I was picturing more of a json format where each county had it's own object. More like:
```
state_names <- unique(wu_data_15_simple$STATE)
wu_list_for_json <- lapply(state_names, function(s) {
    state_df <- dplyr::filter(wu_data_15_simple, STATE == s)
    county_names <- unique(state_df$COUNTY)
    state_list <- lapply(county_names, function(c) {
        county_df <- dplyr::filter(state_df, COUNTY == c)
        return(county_df)
    })
    names(state_list) <- county_names
    return(state_list)
})
names(wu_list_for_json) <- state_names
write(jsonlite::toJSON(wu_list_for_json, pretty=TRUE), file=viz[["location"]])
```
Which looks like:
```
{
  "AL": {
    "Autauga County": [
      {
        "STATE": "AL",
        "STATEFIPS": "01",
        "COUNTY": "Autauga County",
        "COUNTYFIPS": "001",
        "FIPS": "01001",
        "YEAR": "2015",
        "countypop": 55.347,
        "total": 56.41,
        "thermoelectric": 11.86,
        "publicsupply": 3.64,
        "irrigation": 3.52,
        "industrial": 34.8
      }
    ],
    "Baldwin County": [
      {
        "STATE": "AL",
        "STATEFIPS": "01", 
...
```
@aappling-usgs curious how you were picturing it.

viz.yaml
ldecicco-USGS: Actually....I think jsonlite is preferred....old habits die hard...I'll switch over

viz.yaml
ldecicco-USGS: yup

ldecicco-USGS:
Presumably, we can pull both the state and state fip out of this file to save a bunch of data that the user needs to download. ....Should I also split up the json into 51 files (states + PR)? Does that help mobile users? @lindsaycarr @aappling-usgs

lindsaycarr:
I think we need the state info so that data can still show up when we zoom to state-only views? Not if splitting files up is necessary. Let's not worry about that for now and think of that after March 15 when we work on performance issues.

ldecicco-USGS:
So with the update, you'll see in target/data: something like:

```
{"AL":{"Autauga County":[{"COUNTY":"Autauga County","COUNTYFIPS":"001","FIPS":"01001","countypop":55.347,"total":56.41,"thermoelectric":11.86,"publicsupply":3.64,"irrigation":3.52,"industrial":34.8}],"Baldwin County":[{"COUNTY":"Baldwin County","COUNTYFIPS":"003","FIPS":"01003","countypop":203.709,"total":84.7,"thermoelectric":0,"publicsupply":23.67,"irrigation":58.28,"industrial":0}],"Barbour County":[{"COUNTY":"Barbour
```
I straight-up removed the YEAR...I assume this is going to just be 2015 from the get-go. If we need that in here still, holla

scripts/process/simplify_wu_data.R
```diff
@@ -7,24 +7,10 @@ process.simplify_wu_data <- function(viz) {
   orig_wu_type_col <- deps[["col_names"]][["orig_names"]]
   new_wu_type_col <- deps[["col_names"]][["new_names"]]
   
-  columns_to_keep <- c("STATE", "STATEFIPS", "COUNTY", "COUNTYFIPS", "FIPS", "YEAR") # also double as character cols
+  columns_to_keep <- c("STATE", "STATEFIPS", "COUNTY", "COUNTYFIPS", "FIPS", "YEAR")
   columns_to_rename <- c("TP-TotPop", orig_wu_type_col)
   new_names <- c("countypop", new_wu_type_col)
   
-  # none of the columns we kept had characters (-- or N/A) that caused the entire column to
```
lindsaycarr: wait, I think you need to pull from upstream. All of this stuff got deleted.

ldecicco-USGS: I think I reverted some change I made in that file...but that wiped out my upstream pull....so now I need to figure out how to un-revert

lindsaycarr:
Woop! Merging. @ldecicco-USGS please add your estimate on hours spent to the issue, then close the issue!

Issue #14, opened 2018-02-26
URL hash tags


aappling-usgs:
URL should change depending on the view location (national or a state) and category of water use. For example, a URL might look like
owi.usgs.gov/vizlab/water-use-15#st=TX&cat=irr
and typing that URl into a fresh browser tab should bring you back to the same view

aappling-usgs:
Completed with #42 in about 2 hours

Issue #41, opened 2018-02-27
Remove dynamic time features


aappling-usgs:
remove timeslider, updateYear function, activeYear variable

lindsaycarr:
Resolved by #41. Took 20 min.

PR #42, opened 2018-02-27
Url hash tags

aappling-usgs:
Added URL hash tags for water use category. Tags for state were already there. Updated the map title function to reflect current category as well as state.

The default is USA, Total water use:
180227-url-default

You can call the `updateCategory` function in the browser to change the category in both the URL bar and the map title. (This is where @mwernimont can plug in the button functionality - buttons should call `updateCategory()`.)
180227-function-call-zoom

And you can edit the URL directly, or paste in a new URL (not shown) to go to the specified state/national view and water use category:
180227-url-edit-zoom

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -40,7 +40,7 @@ d3.queue()
 var years = [1950, 1960, 1965, 1970, 1980, 1990, 1995, 2000, 2005, 2010, 2015];
 
 // Zoom status: default is nation-wide
-var activeView = getHash('v');
+var activeView = getHash('view');
 if(!activeView) activeView = 'USA';
 
 // Zoom year: start at 2015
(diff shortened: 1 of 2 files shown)


aappling-usgs:
This PR contains a pretty small chunk of javascript code and one small D3 call. @ldecicco-USGS , you interested in reviewing for the sake of more exposure to the .js side? Otherwise I nominate @lindsaycarr 

getHash and setHash are defined in js/url_hashes.js. They read/set the values from URL keys ("v" and "c" in this case).

`d3.select(...).text(...)` allows you to set the value of text in an html item.

I'm creating/using three global variables: activeView, activeYear, and activeCategory. (activeYear to be deleted with #41). These store the current vizzy state and are updated within updateView, updateCategory, etc. in js/map_utils.js.

aappling-usgs:
Resolves #14 . Has taken ~2 hours so far.

aappling-usgs:
done :point_up: and back to you, @lindsaycarr

lindsaycarr:
👍 should I merge or wait for Laura to review also?

aappling-usgs:
go ahead and merge, please.

PR #43, opened 2018-02-27
Static time

aappling-usgs:
This PR should look a lot simpler once https://github.com/USGS-VIZLAB/water-use-15/pull/42 is merged. Just removing the timeslider, activeYear variable, and updateYear function because we're going 2015 all the way in this vizzy now. Resolves #41, which has taken about 20 min so far.

PR #44, opened 2018-02-27
Static time

### LINKED ISSUE #41: Remove dynamic time features

aappling-usgs:
Just removing the timeslider, activeYear variable, and updateYear function because we're going 2015 all the way in this vizzy now. Resolves #41, which has taken about 30 min so far. This is a retry of PR #43


aappling-usgs:
Also renamed `add_states` to `addStates` in .js to conform with https://github.com/USGS-VIZLAB/vizlab/wiki/Style-guide#naming-conventions bullet 7.

js/counties.js
```diff
@@ -36,7 +36,7 @@ function displayCountyData(error, activeCountyData) {
     // extract the topojson to geojson
     var geojson = topojson.feature(activeCountyData, activeCountyData.objects.state);
     var yeargeo = geojson.features.filter(function(d) {
```
lindsaycarr: question I put on your previous PR still stands - should we change the way the county data is processed so that it only includes 2015? Currently preprocessing does it for multiple years, right?

aappling-usgs: oh! yes. i will do that.

lindsaycarr:
Merging and more complicated cleanup will take place with #48.

Issue #32, opened 2018-02-26
Explore pie charts for total category


aappling-usgs:
Could be cool! We'd like to try it, but if they don't look good or don't perform well, we won't use these pie charts in the final product.

aappling-usgs:
1st step: static prototype to see how it looks

aappling-usgs:
2nd step: react to prototype (see #45)

Jordan says: This ended up looking like what I hoped it would - where you get a national view of all of the categories at once, and you can see them stratify spatially. I think it is kind of cool, but I also think it has the potential to overcomplicate what could be a really simple/sharp viz.

Alison says: What about including this view as "Everything" but also having a "Total" view that's just blue circles? I think this one is aesthetically pleasing, which is the first requirement for a successful vizzy. Maybe we should ask some outsiders to give a reaction too, but my first guess is that with both Everything and Total views, the overall viz will be sufficiently simple/clear.

Any other general reactions, ideas for tweaks, opinions on whether to go forward?

aappling-usgs:
Jordan has spent 2 hrs so far

jread-usgs:
Agreed with "everything" and "total". 

An option here too is if the transition between pie to category circle is too heavy for browsers, we could depart from it for the viz and animate the transition for the twitter/reddit gif instead. Noting the risk of implementing this in the viz - if it fails, we still may have a use for it in gif-space.

aappling-usgs:
* Nancy and Cheryl really liked it. We mostly love it too, but probably just because we're nerds
* can be misinterpreted as additional dots on the map - would need extra, careful explanation
* likely to bomb on mobile
* maybe more effective to use as a key story? maybe simplify to thermo vs ag with slider? 
* do we need a focus group to tell us about this?
* remember that this info is largely redundant with what we're already presenting
* we'd need to deal with the gray slices (Other category)
* offer this (static version) as a cool graphic to be used by the WU team as they want - e.g., for them to use in a news release
* write up as a blog? could happen after the viz

gut group response - we're leaning toward not using this in the interactive vizzy...but don't burn the code!!

jread-usgs:
I think it is worth mocking up a gif animation to show what the pie -> category transition would look like. At a minimum, we could provide it to the WU team for future ppt presentations. Also could be part of the social campaign. 

I agree that there is a small downside and risk to including it in the viz - both because of challenges with user interpretation and the practical challenge of writing a performant transition between pie and category.

[...]

PR #45, opened 2018-02-28
first step in #32

jread-usgs:
2 hrs. makes this:
circle_map_total

### DIFF
--- a/scripts/visualize/make_circle_map.R
+++ b/scripts/visualize/make_circle_map.R
@@ -21,9 +21,9 @@ visualize.make_circle_map <- function(viz){
   
   cols <- color_by_wu_type(wu_type)
   
-  png(viz[["location"]], width = 1200, height = 800, res=100)
+  png(viz[["location"]], width = 10, height = 6.0, res=450, units = 'in')
 
-  par(mar = c(0,0,0,0))
+  par(mai=c(0,0,0,0), omi=c(0,0,0,0), xaxs = 'i', yaxs = 'i')
   plot(sf::st_geometry(map_data_usa), col = "#f1f1f1", border = "white")
   plot(sf::st_geometry(map_data_county), col = NA, border = "white", add=TRUE)
   points(circle_sp_transf, pch=21, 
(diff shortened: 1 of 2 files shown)


scripts/visualize/make_circle_map.R
```diff
@@ -33,6 +33,76 @@ visualize.make_circle_map <- function(viz){
...
+  
+  for (j in seq_len(length(circle_sp_transf))){
+    # this is in a loop because it is throwaway code: 
+    r <- circle_sp_transf$radius_total[j] *rad_multiplier
+    c.x <- coordinates(circle_sp_transf)[j, ][['x']]
+    c.y <- coordinates(circle_sp_transf)[j, ][['y']]
+    for (cat in categories){
+      cat_angle <- circle_sp_transf[[cat]][j] / circle_sp_transf[['total']][j]*2*pi
+      if (cat == head(categories, 1L)){
+        # start the first category mirrorer relative to the top
```
aappling-usgs: mirrorer?

scripts/visualize/make_circle_map.R
```diff
@@ -33,6 +33,76 @@ visualize.make_circle_map <- function(viz){
...
+  
+  
+  png(viz[["location"]], width = 10, height = 6.0, res=450, units = 'in')
+  
+  par(mai=c(0,0,0,0), omi=c(0,0,0,0), xaxs = 'i', yaxs = 'i')
+  plot(sf::st_geometry(map_data_usa), col = "#f1f1f1", border = "white")
+  plot(sf::st_geometry(map_data_county), col = NA, border = "white", add=TRUE, lwd = 0.5)
+  
+  # start w/ irrigation, fill around the categories (make sure they add up!)
+  # note: they don't add up because we aren't including all
```
aappling-usgs: my guess is we'll want an "Other" category if we decide to include this in the final viz

scripts/visualize/make_circle_map.R
jread-usgs: geez, terrible comment by me. What I meant was the first category (in this case) will have a slice centered on the 90° like this:
image

scripts/visualize/make_circle_map.R
jread-usgs: agreed

scripts/visualize/make_circle_map.R
aappling-usgs: oh, now i see where the mirror concept comes in :+1:

PR #47, opened 2018-02-28
update required packages in viz.yaml

lindsaycarr:
Added these because I had a fresh library so it error-ed every single time.


lindsaycarr:
✋ guilty, too

Issue #26, opened 2018-02-26
Embeddable figures


aappling-usgs:
The national and state views should be embeddable, which I think means
(1) a figure-only html page should be available for each view, and
(2) there should be a button / buttons on the main page that provide the embeddable links

lindsaycarr:
Related to https://github.com/USGS-VIZLAB/vizlab/issues/188 ?

lindsaycarr:
Closed by #54. Took 1 hr.

PR #49, opened 2018-03-01
Adding social icons (pre-embed pr)

ldecicco-USGS:
Took ~ 1 hour (rounding up....I'll probably round down on the next PR to compensate). I want to get this merged and finished before *really* tackling the embed stuff. Here's what it looks like:

image

### DIFF
--- a/viz.yaml
+++ b/viz.yaml
@@ -79,6 +79,12 @@ parameter:
     id: wu_type_column_names
     orig_names: ["TO-Wtotl", "PT-Wtotl", "PS-Wtotl", "IR-WFrTo", "IN-Wtotl"]
     new_names: ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"]
+  - 
+    id: social-links
+    github: https://github.com/USGS-VIZLAB/water-use-15
+    twitter: https://twitter.com/share?url=TODO!!!
+    facebookLink: http://www.facebook.com/sharer.php?u=TODO!!!
+  
 
 fetch:
   -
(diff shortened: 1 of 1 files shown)


viz.yaml
```diff
@@ -265,11 +271,14 @@ publish:
     template: fullpage
     publisher: page
     depends:
+      sourcesansprofont: sourcesansprofont
```
ldecicco-USGS: I'm mucking around with this to try to figure out why the footer looks dumb again....I don't think this is it...I'll remove

viz.yaml
```diff
@@ -79,6 +79,12 @@ parameter:
     id: wu_type_column_names
     orig_names: ["TO-Wtotl", "PT-Wtotl", "PS-Wtotl", "IR-WFrTo", "IN-Wtotl"]
     new_names: ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"]
+  - 
+    id: social-links
+    github: https://github.com/USGS-VIZLAB/water-use-15
+    twitter: https://twitter.com/share?url=TODO!!!
+    facebookLink: http://www.facebook.com/sharer.php?u=TODO!!!
```
lindsaycarr: let's not forget "TODO!!!" these 😄

lindsaycarr:
So @ldecicco-USGS is this time linked to #26 stuff or separate?

ldecicco-USGS:
This is basically pre-#26...trying to keep the PR to a reasonable length, seemed like doing this obvious work first would help. (I need to do this for embed....but if somehow the embed work completely fails because I'm confused on url hash stuff...at least we'll still have the basics)

PR #51, opened 2018-03-01
Buttons for switching categories

mwernimont:
So I know there could be a bunch improved here, which I expect you guys will let me know. Sorry for the tempData thing, I had no clue where to access any data to dynamically make the buttons, so just made a temp DB to show what they will look like until something better and more sustainable can replace it.

I also am running with an idea to switch the buttons into a select input for mobile, as the buttons don't look great on mobile stacked or next to each other. Let me know what you guys think.  I can be veto'd no problem.

Currently, only the buttons are hooked up to the updateCategory() function, since the select was giving me issues, and until we have a consensus that its a good idea, didn't want to waste more time on it.

Spent 4 hours on this.


layout/css/main.css
```diff
@@ -0,0 +1,102 @@
+*{
+  padding:0;
+  margin:0;
+}
+
+body {
+  font-family: sans-serif;
+}
+
+#buttonContainer{
```
aappling-usgs: can we do kebab-case for these to work with the style guide https://github.com/USGS-VIZLAB/vizlab/wiki/Style-guide?

js/build_map.js
```diff
@@ -9,16 +9,21 @@ var projection = albersUsaPr()
     .translate([chart_width / 2, chart_height / 2]);
 var buildPath = d3.geoPath()
     .projection(projection);
+    
+//Create container
+var container = d3.select('body')
```
aappling-usgs: what does the container contain? could it be named to reflect the contents?

js/build_map.js
```diff
@@ -9,16 +9,21 @@ var projection = albersUsaPr()
...
   .range([0, 20])
 
 // Create SVG
-var svg = d3.select("body")
+var svg = d3.select("#container")
     .append("svg")
-    .attr("width", chart_width)
-    .attr("height", chart_height);
+    .attr('viewBox', '0 0 ' + chart_width + ' ' + chart_height + '')
+	  .attr('preserveAspectRatio', 'xMidYMid');
```
aappling-usgs: is this indentation intended?

js/build_map.js
mwernimont: Container contains everything.  The parent of all besides the footer and Header. I can rename it.

js/build_map.js
mwernimont: Nope

mwernimont: But does seem to be a GitHub thing, cause I don't have that indent in my code

aappling-usgs: hmm. tabs versus spaces? i think javascript people might be more tab-centric, but lindsay and i have been tending toward spaces because RStudio will autoconvert tabs to spaces for you and that's the recommended style for R code...

mwernimont:
CSS elements are now - instead of camelCase.  #container, is now 
3content-container.  Let me know if the specificity isn't specific enough.

PR #54, opened 2018-03-01
Adding embed icon and ability to embed.

ldecicco-USGS:
Total time so far: 1 hr.

image

that shows the embed link.....
Here's what you get when you click it:

http://127.0.0.1:4321/national-water-use.html

image

I think all the url-hash stuff works with that embed feature...and...I...think(?) that's good...but not 100% sure.

### DIFF
--- a/viz.yaml
+++ b/viz.yaml
@@ -360,10 +360,10 @@ publish:
     publisher: social
     depends: social-links
     context:
-      embedLink: national-water-use.html
+      embedLink: embed-water-use-15.html
   -
     id: national-water-use
-    name: national-water-use
+    name: embed-water-use-15
     publisher: page
     template: embed
     mimetype: text/html
--- a/viz.yaml
+++ b/viz.yaml
@@ -84,7 +84,6 @@ parameter:
     github: https://github.com/USGS-VIZLAB/water-use-15
     twitter: https://twitter.com/share?url=TODO!!!
     facebookLink: http://www.facebook.com/sharer.php?u=TODO!!!
-  
 
 fetch:
   -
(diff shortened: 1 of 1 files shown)


lindsaycarr:
I think it's good that the hash stuff is carried through.

ldecicco-USGS:
Yeah, I think that just gives us the state embed functionality we were looking for. I just wasn't sure I've thought through all the implications...

ldecicco-USGS:
or, you could embed a link that starts with irrigation in California or something

lindsaycarr:
Even if someone embeds something for just Texas and irrigation, people should be able to interact with it so that they can see the whole country for total. I think that is the behavior we want!

viz.yaml
```diff
@@ -360,5 +359,32 @@ publish:
     template: vertical-social-media-template
     publisher: social
     depends: social-links
-
+    context:
+      embedLink: national-water-use.html
```
lindsaycarr: do we want that to be "national-water-use" or "2015-water-use"?

ldecicco-USGS: Let's see...the base url is going to be:
https://owi.usgs.gov/vizlab/water-use-15
maybe we want something more simple like:
https://owi.usgs.gov/vizlab/water-use-15/embed.html
?
I guess I don't have a great opinion one way or another...for reference we did it in Irma like this:
https://owi.usgs.gov/vizlab/hurricane-irma/embed-hurricane-irma.html

lindsaycarr: hmmm OK, let's use just `embed` then? Would the hashes end up before or after the word?

aappling-usgs: or embed-water-use-15?

ldecicco-USGS: https://owi.usgs.gov/vizlab/water-use-15/embed-water-use-15.html
?
I'll push that up in a sec......unless I hear otherwise in the meantime....

aappling-usgs: the hashes _should_ end up after the word. say something if they don't.

either embed-water-use-15.html or embed.html works for me.

ldecicco-USGS: Yeah, they do. So, if you type into your browser:
https://owi.usgs.gov/vizlab/water-use-15/embed-water-use-15.html
It automatically adds the hash stuff.
We can also directly add the hash stuff to a particular state/category for a particular news outlet. So....I'm pretty sure we're all on the same page...I just *wasn't* sure we would be before I opened this PR.

ldecicco-USGS:
Bumping this up to 2 hours

Issue #19, opened 2018-02-26
Improve tooltip performance


aappling-usgs:
How can users access overlapped circles? Are there speed issues to be addressed? Do the tooltips always look their best?

lindsaycarr:
Initial circle opacity is slowing this down. If CSS color definitions for circles had alpha built in, it might help.

lindsaycarr:
Think about hover capability for tiny tiny circles that are inside larger circles. Can people successfully hover? See https://github.com/USGS-VIZLAB/water-use-15/pull/55#discussion_r171679881. Also, think about hovers on mobile...

jread-usgs:
See references for profiling: https://developers.google.com/web/tools/chrome-devtools/evaluate-performance/reference

[Current version](https://github.com/USGS-VIZLAB/water-use-15/pull/57) is maxing CPU whenever we move the mouse on a circle 
![image](https://user-images.githubusercontent.com/2349007/36904616-c277b4e0-1df6-11e8-863f-5ea12fff260d.png)

Very little CPU is used for anything other than Rastering

jread-usgs:
Much more manageable w/ opaque circles (but that along isn't a solution for us):

![image](https://user-images.githubusercontent.com/2349007/36905128-237e85f6-1df8-11e8-94f3-690b755871a4.png)

jread-usgs:
and it looks like @lindsaycarr is right, baking the alpha into the fill instead of opacity makes this run just as well as (or at least I couldn't tell the difference) 100% opaque:

![image](https://user-images.githubusercontent.com/2349007/36905250-88cca0a0-1df8-11e8-9350-68839bfbe545.png)

aappling-usgs:
cool trick!

aappling-usgs:
Where are the dots on the right edge in that example? you're not talking about the legend, are you?

wdwatkins:
The thermoelectric circles to the east of Oregon.

aappling-usgs:
Sounds like removing sorting and hovering on counties rather than dots is solving this sufficiently for us (done in #106). We can add a new issue later if we find that there's more to be done.

PR #55, opened 2018-03-01
Tooltips

lindsaycarr:
What this adds: tooltip with info about county name, population, and water use value. Circle that is being hovered on changes to opacity 1 (rather than 0.8) and is brought to the front. When the mouse moves away, the circle goes back to it's original position (functionally, a duplicate is being made and then deleted). For some reason, the hovers became slow when I added an initial opacity (commented that in the line).

For now this is the rest of #35. Performance can be worked on with #19.

image

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -17,12 +17,11 @@ var container = d3.select('body')
 
 // circle scale
 var scaleCircles = d3.scaleSqrt()
-  .range([0, 20])
+  .range([0, 20]);
 
 // Setup tooltips
 var tooltipDiv = d3.select("body").append("div")
-      .attr("class", "tooltip")
-      .style("opacity", 0);
+      .classed("tooltip hidden", true);
 
 // Create SVG
 var svg = d3.select("#content-container")
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -14,6 +14,11 @@ var buildPath = d3.geoPath()
 var scaleCircles = d3.scaleSqrt()
   .range([0, 20])
 
+// Setup tooltips
+var tooltipDiv = d3.select("body").append("div")
+      .attr("class", "tooltip")
+      .style("opacity", 0);
+
 // Create SVG
 var svg = d3.select("body")
     .append("svg")
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -2,6 +2,9 @@
 var chart_width     =   1000;
 var chart_height    =   700;
 
+// define categories
+var tempCategories = ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"];
+
 // Projection
 var projection = albersUsaPr()
     .scale([1200])
@@ -49,9 +52,6 @@ d3.queue()
   .defer(d3.json, "data/county_centroids.json")
   .await(create_map);
 
-// dummy var for now
-var years = [1950, 1960, 1965, 1970, 1980, 1990, 1995, 2000, 2005, 2010, 2015];
-
 // Zoom status: default is nation-wide
 var activeView = getHash('view');
 if(!activeView) activeView = 'USA';
@@ -96,8 +96,6 @@ function create_map() {
 
 }
 
-var tempCategories = ["Total", "Thermoelectric", "Public Supply", "Irrigation", "Industrial"];
-
 var buttonContainer = d3.select('#content-container')
   .append('div')
   .attr('id', 'button-container');
@@ -126,11 +124,10 @@ var categoryButtons = d3.select('#button-container')
   .enter()
   .append('button')
   .text(function(d){
-    return d;
+    return categoryToName(d);
   })
   .attr('class', function(d){
-    var name = d.split(' ');
-    return name[0];
+    return d;
   })
   .on('click', function(d){
     updateCategory(d.toLowerCase());
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -64,7 +64,7 @@ function addCentroids(map, countyCentroids) {
     .attr("r", function(d) { 
       return scaleCircles(d.properties[[activeCategory]]);
     })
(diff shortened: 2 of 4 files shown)


js/map_utils.js
```diff
@@ -64,8 +64,10 @@ function addCentroids(map, countyCentroids) {
     .attr("r", function(d) { 
       return scaleCircles(d.properties[[activeCategory]]);
     })
+    .on("mouseover", function(d) { seeTooltip(this, d); })
+    .on("mouseout", function(d) { hideTooltip(this, d); })
```
aappling-usgs: a question about these tooltips: they're currently attached to the circles, and i like what you're doing with bringing the hovered-over circle to the top. but how does and should this hover behavior interact with hovering over a county polygon? should/can we add the same hover functions to the polygons when those are ready? looking for your educated opinion here.

lindsaycarr:
Oops - not actually using "moveToBack" anymore...I need to delete that. Trying out the hovertext function from vizlab first - https://github.com/USGS-VIZLAB/vizlab/blob/d38ee8d61d060f7e6da2d6f4fe930cbb7dc773c0/inst/js/tooltip.js

js/map_utils.js
```diff
@@ -258,3 +260,77 @@ function updateCircles(activeCategory) {
...
+    .duration(100);
+}
+
+d3.selection.prototype.moveToFront = function() {  
+      return this.each(function(){
+        this.parentNode.appendChild(this);
+      });
+    };
+    
+d3.selection.prototype.moveToBack = function() {
```
aappling-usgs: does moveToBack get used?

js/map_utils.js
```diff
@@ -258,3 +260,77 @@ function updateCircles(activeCategory) {
...
+    return this.each(function() { 
+      console.log(this.parentNode);
+        var firstChild = this.parentNode.firstChild; 
+        if (firstChild) { 
+            this.parentNode.insertBefore(this, firstChild); 
+        } 
+    });
+};
+
+function categoryToClass(category) {
```
aappling-usgs: this is getting to be a lot of mappings. do we need them all?

js/map_utils.js
```diff
@@ -258,3 +260,67 @@ function updateCircles(activeCategory) {
...
+    .duration(100);
+}
+
+d3.selection.prototype.moveToFront = function() {  
+      return this.each(function(){
+        this.parentNode.appendChild(this);
+      });
+    };
+
+function categoryToClass(category) {
```
aappling-usgs: this is getting to be a lot of ways of expressing category. can we keep it to one internal representation (classes, hash tags, etc.) and one we show the world?

lindsaycarr: ^ yes I would like that. @mwernimont can we update the classes to be what is in the data already:
```
total
thermoelectric
publicsupply
irrigation
industrial
```

mwernimont: For the buttons?

js/map_utils.js
lindsaycarr: 🤔 thinking ....

lindsaycarr: I think it could be easily done if I pull 

```
var orig = d3.select(currentCircle),
      origNode = orig.node();
  var duplicate = d3.select(origNode.parentNode.appendChild(origNode.cloneNode(true), 
                                                            origNode.nextSibling));
  
  // style circles
  orig
    .style("opacity", 0); // makes original circle invisible in the background
  duplicate
    .classed('county-point-duplicate', true)
    .style("pointer-events", "none")
    .style("opacity", 1); // makes the duplicate circle on the top
```
out to the `.on("mouseover"` specifically for the circle. Then we could do `.on("mouseover", function(d) { showTooltip(d); })` where ever the counties are added. The only thing that I don't think would work is that the `d` used here when it's associated with a circle has the actual data, but it wouldn't for the county boundary.

aappling-usgs: ok, that sounds like a good approach. let's hold off until another PR regardless, but maybe what we need to do is index both the county topo and the centroid data on GEOID so we can quickly cross reference.

aappling-usgs: what happens if there's a little circle that ends up getting completely swamped by a big circle? i imagine that as you move your mouse toward the little circle, it hits the big circle and brings that one to the front, such that it's actually impossible to get the little circle to show. do you think that's accurate? and does it happen at all / often enough to be a problem worth our attention?

js/map_utils.js
lindsaycarr: Marty and I talked. Yes, we would like to try to get the class to be the same, but realized that we need probably need multiple class definitions anyways since buttons need the color defined by `background` and the circles need `fill`

aappling-usgs: makes sense. so what's your plan?

js/map_utils.js
lindsaycarr: hmmm that's a good question. Doesn't appear to be a problem because I turned pointer events off for those duplicated circles, so there isn't any "hover" capability turned on when those are on top of everything else. Good thing to look into more though. Can we consider that with tooltip performance so that we can actually merge this and have people play around on dev?

lindsaycarr: Here's the tiny dot I tested it on in california:
image

lindsaycarr: I could hover over it (mobile might be a different story though)

aappling-usgs: cool, thanks for checking! yes, absolutely i'll merge soon, no refinement needed yet. just trying to figure out which refinements/planning will be needed.

js/build_map.js
```diff
@@ -126,11 +124,10 @@ var categoryButtons = d3.select('#button-container')
   .enter()
   .append('button')
   .text(function(d){
-    return d;
+    return categoryToName(d);
```
aappling-usgs: :+1:

PR #56, opened 2018-03-01
Projection fix

mwernimont:
screen shot 2018-03-01 at 3 54 22 pm
Should get the dots showing up and on top of the states.

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -6,9 +6,10 @@ var chart_height    =   700;
 var tempCategories = ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"];
 
 // Projection
-projection.scale([1200])
-    // default is .rotate([96,0]) to center on US (we want this)
+// Projection
+var projection = albersUsaTerritories().scale([1200])
     .translate([chart_width / 2, chart_height / 2]);
+    // default is .rotate([96,0]) to center on US (we want this)
     
 var buildPath = d3.geoPath()
     .projection(projection);
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -3,7 +3,7 @@ var chart_width     =   1000;
 var chart_height    =   700;
 
 // Projection
-var projection = albersUsaPr()
+var projection = albersUsaTerritories()
     .scale([1200])
     // default is .rotate([96,0]) to center on US (we want this)
     .translate([chart_width / 2, chart_height / 2]);
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -6,10 +6,10 @@ var chart_height    =   700;
 var tempCategories = ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"];
 
 // Projection
-var projection = albersUsaTerritories()
-    .scale([1200])
+projection.scale([1200])
     // default is .rotate([96,0]) to center on US (we want this)
     .translate([chart_width / 2, chart_height / 2]);
+    
 var buildPath = d3.geoPath()
     .projection(projection);
     
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -47,7 +47,35 @@ function albersUsaTerritories() {
         t = lower48.translate(),
         x = (coordinates[0] - t[0]) / k,
         y = (coordinates[1] - t[1]) / k;
-        
+        /*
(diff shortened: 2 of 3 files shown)


js/custom_projection.js
```diff
@@ -1,146 +1,290 @@
...
+        console.info("p0 guam", x0 + ' - ' + y0);
+        c1 = guam(guamBbox[1]);
+        x1 = (c1[0] - t[0]) / k;
+        y1 = (c1[1] - t[1]) / k;
+        console.info("p1 guam", x1 + ' - ' + y1);
+        */
+
+    return (y >= 0.120 && y < 0.234 && x >= -0.425 && x < -0.214 ? alaska
+        : y >= 0.166 && y < 0.234 && x >= -0.214 && x < -0.115 ? hawaii
+        : y >= 0.2064 && y < 0.2413 && x >= 0.312 && x < 0.385 ? puertoRico
```
aappling-usgs: are PR & VI handled together?

js/custom_projection.js
```diff
@@ -1,146 +1,290 @@
...
   };
 
-  return albersUsa.scale(1000);
+
+  return albersUsa.scale(1070);
+}
+
+// The projections must have mutually exclusive clip regions on the sphere,
+// as this will avoid emitting interleaving lines and polygons.
+function multiplex$2(streams) {
```
aappling-usgs: does multiplex$2 get used?

js/map_utils.js
```diff
@@ -33,7 +33,9 @@ var stateStyle = {
 
 function addCentroids(map, countyCentroids) {
   
-  var tempProjection = d3.geoAlbersUsa();
+  var tempProjection = albersUsaTerritories()
```
aappling-usgs: can we use just one projection now? there's `projection` in build_map.js. i only created `tempProjection` so i could get some dots in there for the prototype without completely breaking the vizzy. ideally we'd just assign `albersUsaTerritories` to `projection`, delete `tempProjection`, and use `projection` everywhere

jread-usgs:
Can you put a screenshot in the PR?

js/custom_projection.js
mwernimont: So that is what I need the actual shapes for, I can adjust and do what we need to get both, but until then I am just guessing.

mwernimont: But I feel 95% sure on positivity they will be

js/map_utils.js
mwernimont: If you know how to get that done smoothly, I'm all for one projection.

js/custom_projection.js
aappling-usgs: cool

js/map_utils.js
aappling-usgs: i think it's just doing what i wrote above: delete the definition of tempProjection, define projection = albersUsaTerritories, and replace any uses of tempProjection with projection.

mwernimont:
@aappling-usgs removed that multiplex$2 as you were right, it wasn't used.  And tempProjection is now projection.

mwernimont:
So unless my browsers were cached it worked...

js/build_map.js
```diff
@@ -6,10 +6,10 @@ var chart_height    =   700;
 var tempCategories = ["total", "thermoelectric", "publicsupply", "irrigation", "industrial"];
 
 // Projection
-var projection = albersUsaTerritories()
-    .scale([1200])
+projection.scale([1200])
```
aappling-usgs: you should be able to make just one projection.scale() call, too - i think you can take out this call here, only define the projection and projection details within map_utils.js. but could you do me a favor and copy over the comment ` // default is .rotate([96,0]) to center on US (we want this)` to map_utils.js? took me hours to figure that out =)

lindsaycarr:
I'm going to let alison take this review. Haven't been part of the projection stuff at all this go around - I'll let you two hash it out :)

aappling-usgs:
looks like that should do it :+1:

lindsaycarr:
Not totally done because of PR and VI verification needed still, but significant progress towards closing #9

aappling-usgs:
Yes, significant! :tada:

Issue #48, opened 2018-02-28
Clean up geospatial datasets


aappling-usgs:
We have 3 different sources of spatial data right now:
1) state boundaries from the `maps` package
2) county boundaries, including historical ones but excluding the virgin islands, from NHGIS IPUMS. these are the ones that are handled by preprocessing.yaml.
3) county boundaries, 2015 only, straight from the census website, and including puerto rico and the virgin islands

I think we really just need dataset 3, and from that dataset we need:
* county boundaries for the whole US all at once, probably not split into one file per state unless/until we learn that's necessary for performance.
* county centroids computed from the same dataset, possibly merged with the water use data.
* state centroids for the purpose of zooming
* state boundaries? do we need them any more?

For this issue, I propose that we remove datasets 1 and 2 and then use dataset 3 to meet the needs stated above. If we can get away without state boundaries, we should also stop generating/saving those.

aappling-usgs:
I'm at 6.5 hours as of submitting #57

aappling-usgs:
closed by #75

PR #57, opened 2018-03-01
Simpler spatial

aappling-usgs:
This PR takes a step or two toward #48. Here I have:
* removed the alternate county data sources that we weren't using at all
* constructed a national topojson file of county boundaries starting from the Census Bureau's data (approved by the WU team)
* reduced reliance on the IPUMS-based files by displaying Census Bureau data rather than IPUMS data as county boundaries in the vizzy

In a future PR or two we should:
* finish eliminating reliance on the IPUMS-based files by generating stateDict and countyDict from the Census Bureau files instead
* after that, we can either take the preprocessing step out or push the census boundaries data manipulation into preprocessing
* generate state boundaries by merging counties rather than by querying `maps` (this will improve the match between state and county boundaries)
* shift reliance from the current state boundaries to those new merged-county boundaries

And when we get to performance refinements we should consider
* keeping the 1 big national file of county boundaries, but maybe simplifying it to the minimum needed for a national view. then also build state-specific files with more detailed county boundaries, and only load those files on demand. build on the existing state-data-caching code in map_utils.js


lindsaycarr:
@mwernimont here's your answer to where she's at with the topojson :)

viz.yaml
```diff
@@ -109,19 +103,23 @@ fetch:
...
     scripts: []
-    # this file includes counties for Northern Mariana Island (STATEFP=='69'), Guam (STATEFP=='66'), and American Samoa (STATEFP=='60')"
-    # and also the territories we do want: Virgin Islands and Puerto Rico"
+    # see https://www.census.gov/geo/maps-data/data/cbf/cbf_counties.html for details.
+    # this file includes counties for places we don't need:
+    #   Northern Mariana Island (STATEFP=='69'), Guam (STATEFP=='66'), and American Samoa (STATEFP=='60')
+    # and also the places we do want:
+    #   Virgin Islands, Puerto Rico, and all 50 states
   -
     id: county_boundaries_topojson
```
lindsaycarr: Was there a reason for keeping this?

js/counties.js
```diff
@@ -5,44 +5,77 @@ function hideCounties() {
...
-    var yeargeo = geojson.features.filter(function(d) {
-      return d.properties.year == 2015;
-    });
-  
     // create paths
     var countyBounds = map.selectAll(".county")
-      .data(yeargeo, function(d) {
-        return d.properties.county_FIPS;
+      .data(activeCountyData, function(d) {
+        return d.properties.GEOID;
```
lindsaycarr: thinking ahead, is this where we would easily drop or add circles when zoomed into a state level?

lindsaycarr:
@aappling-usgs is this what allows us to start using PR and VI shapes for @mwernimont's projection task?

aappling-usgs:
not yet on the state shapes, which still currently rely on a dataset that doesn't have PR and VI. next PR, i hope.

but in the meantime, @mwernimont , you should be able to stop filtering out AK, PR, and VI centroid dots, and if the projection doesn't break with those dots added back in, then the projection is likely to work for the state shapes as well once we have them.

js/counties.js
aappling-usgs: yes, i think that's right...

viz.yaml
aappling-usgs: we still need county_boundaries_topojson for now because we're relying on it for the state and county dictionaries. as soon as we're creating those from county_boundaries_topojson3, we can remove this one. and then take the 3 off of county_boundaries_topojson3.

aappling-usgs:
PR and VI dots work!
image

viz.yaml
lindsaycarr: 👍

PR #64, opened 2018-03-02
Preprocess counties

aappling-usgs:
Trying to do the minimum to get the topojson effort out of the viz.yaml and into preprocess.yaml. I think this does it.


aappling-usgs:
Yes, that'll be a different PR - still need to extract the state and county dictionaries before we can get rid of that step.

Issue #38, opened 2018-02-27
Transitions for switching categories


aappling-usgs:
Write the functions that are triggered by the buttons in #21

aappling-usgs:
This issue now includes hooking up the buttons so that they trigger category changes, as well as making those category changes look nice or at least acceptable.

aappling-usgs:
If it's easy, want to spend a few minutes experimenting with also hooking up the buttons so that hovering over a button temporarily displays the hovered-over category (and then reverts when the mouse moves off)? If it's hard, feel free to kick that idea over to #63 . But I think it's just a matter of (1) adding the same function to on('mouseover') that you'll already be adding to on('click'), plus some state tracking so that on('mouseout') can switch back to the most recently clicked category.

jread-usgs:
Seems this is complete?

aappling-usgs:
Good with me. Lindsay said in PR that this took 2 hours.

PR #65, opened 2018-03-02
Button transitions!

lindsaycarr:
This is for #38. This switches the categories when the button is clicked, and allows a hover on the button to "preview" the category, which changes the title, too. Example gif with a mix of button clicks + hovers:

hover_click_change_cats

also @mwernimont it looks like alaska is sitting up their in the top left corner - I see a glimmer of blue. Is that helpful or did you know that already?

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -115,5 +115,12 @@ var categoryButtons = d3.select('#button-container')
     return d;
   })
   .on('click', function(d){
+    activeCategory = d.toLowerCase(); // put this here so it only changes on click
+    updateCategory(activeCategory);
+  })
+  .on('mouseover', function(d){
     updateCategory(d.toLowerCase());
+  })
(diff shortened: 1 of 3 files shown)


lindsaycarr:
This took almost 2 hrs.

js/build_map.js
```diff
@@ -115,5 +115,12 @@ var categoryButtons = d3.select('#button-container')
...
   })
   .on('click', function(d){
+    activeCategory = d.toLowerCase(); // put this here so it only changes on click
+    updateCategory(activeCategory);
+  })
+  .on('mouseover', function(d){
     updateCategory(d.toLowerCase());
+  })
+  .on('mouseout', function(d){
+    updateCategory(activeCategory);
```
aappling-usgs: wow, that's even simpler than i imagined! nice solution

PR #68, opened 2018-03-05
Adding some text

### LINKED ISSUE #36: Add text to vizzy
Follows #23 and #24. Get the draft text in place in the vizzy.

ldecicco-USGS:
Not good text, just the pipings *for* text:

Fixes #36 

Main:
image

Embed (not sure we want text here or not...but if we do, it's here:)

image

### DIFF
--- a/data/description_text.yaml
+++ b/data/description_text.yaml
@@ -1,0 +1,36 @@
+---
+description_text-markdown: |+
+  ## Interesting info about water use
+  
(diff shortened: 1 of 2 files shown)


ldecicco-USGS:
2 hours because fiddling with markdown vs plain text. Doesn't seem to matter much for plain text plus links, but headings, lists, and simple tables will probably be easier via markdown.

lindsaycarr:
Might be the easiest way to add any of those key stories with static figs, too :)

Issue #60, opened 2018-03-02
"Get the code" link in footer seems wrong


ldecicco-USGS:
Currently going back to main link (not github)

lindsaycarr:
This should be a vizlab issue, not a water use viz issue. I noted this awhile back for how-to-gdp https://github.com/USGS-VIZLAB/how-to-gdp/issues/52. I made this vizlab issue https://github.com/USGS-VIZLAB/vizlab/issues/339

lindsaycarr:
If this gets fixed by fixing for all of vizlab through https://github.com/USGS-VIZLAB/vizlab/issues/339, hours won't count towards water-use-viz. A fix exclusively in water-use would. Opening to be able to make sure it happens.

lindsaycarr:
@ldecicco-USGS how long did this take? We will count hours for this guy as just "vizlab".

ldecicco-USGS:
OK, <1 hour

PR #70, opened 2018-03-05
Getting the "Get the Code" link to work

ldecicco-USGS:
Also added a hurricane viz in for now. If we don't fix the footer style, at least it'll look OK.

image


lindsaycarr:
This closes #60.

Issue #74, opened 2018-03-13
clean up prototypes


aappling-usgs:
maybe we leave them in for now, but `centroid_sizes`, `circle_map_total`, `circle_map_thermoelectric`, `circle_map_publicsupply`, `circle_map_irrigation`, and `circle_map_industrial` are all viz items that don't make it into the dynamic viz and use different methods for preparing both spatial and water use data.

ultimately i'd like to see us either streamline the methods (if we find another use for these prototypes) or delete them.

jread-usgs:
Yes, I think they should be removed and I always feel a bit odd about where the prototyping stuff fits into the viz repo. Should it be separate and included in `viz-scratch` in the future? Is that too decoupled or redundant from the data fetching/processing steps?

aappling-usgs:
Not sure - seemed like the right place at the time, but maybe the cleanup effort would have been smaller if we'd kept it separate in `viz-scratch`.

aappling-usgs:
Since you're good with removing, and I'm hitting errors in trying to get them to work again with some recent spatial data changes. I'm going to remove them right now.

aappling-usgs:
closed by #75

PR #75, opened 2018-03-13
Simpler spatial data

aappling-usgs:
* Moved most of the spatial data handling into preprocess (sorta #48)
* Calculated state boundaries, and state and county dictionaries, from the census county boundary dataset; eliminated reliance on 2 other spatial data sources (#48)
* Removed viz fetch/process items that became unnecessary after the refactor (#48)
* Merged state and county dictionary info into their topojson files and stopped sending the state and county dictionaries separately over to the browser (#48)
* Removed prototype visualization items that are no longer necessary (#74)
* Resolved the final bit of #9, which was to fix a specific county whose centroids weren't getting projected. What I did was (1) compute "centroids" using `rgeos::gPointOnSurface` instead of `rgeos::gCentroid`, which I figure has higher likelihood of being in a location I deem believable, and (2) add a check on the D3 side to just plot at 0,0 and give a console message if a specific point is a problem. There don't seem to be any problem points anymore, but if any arose, we'd be ready.

There's lots of tweaking and optimization that can be done from here, but this PR is obviously big enough :grimacing:


js/counties.js
```diff
@@ -92,7 +87,7 @@ function displayCountyData(error, activeCountyData) {
         return d.properties.GEOID;
       })
       .text(function(d) {
-        return d.properties.NAME;
+        return d.properties.COUNTY_LONG;
```
aappling-usgs: i think we now have COUNTY and COUNTY_LONG available to us, and I decided that COUNTY looked better more often. i'll switch this if i can.

js/map_utils.js
```diff
@@ -33,33 +33,40 @@ var stateStyle = {
...
-      return projection(d.geometry.coordinates)[0];
+      var proj = projection(d.geometry.coordinates);
+      if(!proj) {
+        console.log('bad projection:');
+        console.log(d);
+        console.log(projection(d.geometry.coordinates));
+        return 0;
+      } else {
+        return projection(d.geometry.coordinates)[0]; 
+      }
```
aappling-usgs: this check is no longer technically necessary, but i bet it's not taking much processing time, and it could be useful if we go messing with the spatial data again (which i expect we will).

aappling-usgs:
Looks like I have a total of 18.5 hours in this simplifying spatial data task, including #64 and #57

js/counties.js
```diff
@@ -91,9 +86,6 @@ function displayCountyData(error, activeCountyData) {
       .attr('id', function(d) {
         return d.properties.GEOID;
       })
-      .text(function(d) {
-        return d.properties.NAME;
-      })
```
lindsaycarr: I don't remember seeing any text before...was this actually working?

js/build_map.js
```diff
@@ -79,13 +77,12 @@ function create_map() {
 	if (error) throw error;
 
 	// the rest of the indices of arguments are all the other arguments passed in -
-	// so in this case, all of the results from q.await
-	stateData = arguments[1];
-	stateDict = arguments[2];
-	countyDict = arguments[3];
-	countyCentroids = arguments[4];
+	// so in this case, all of the results from q.await. Immediately convert to
+	// geojson so we have that converted data available globally.
```
lindsaycarr: 🙏 woo!

js/map_utils.js
lindsaycarr: So, `projection(d.geometry.coordinates)` was returning NULL, and `!proj` is TRUE for NULL? This isn't a critique of your implementation, I haven't used this syntax before so I wanted to understand it.

lindsaycarr: Actual review: would it be more useful to print `d.geometry.coordinates` than just `d`?

scripts/preprocess/process_compute_centroids.R
```diff
@@ -0,0 +1,22 @@
+process_compute_centroids <- function(outfile, spdat, county_dict_file) {
+  
+  # compute the centroids (actually, return a point on the surface of each multipolygon)
+  centroids <- rgeos::gPointOnSurface(spdat, byid=TRUE, id=spdat@data$GEOID)
```
lindsaycarr: can you just add a note about how this important for groups of islands compared to `gCentroid`? I want to remember that for the future, but I have a terrible memory and prefer things to be written :)

scripts/preprocess/process_compute_centroids.R
```diff
@@ -0,0 +1,22 @@
...
+  
+  # edit the properties in the spdat, removing unnecessary ones and adding
+  # others from county_dict. similar to process_counties.R
```
lindsaycarr: I'm a little confused about "remove unnecessary ones". Are there examples? How do you which data is the authoritative county list, `county_dict_file` or `spdat`?

scripts/preprocess/process_dictionaries.R
```diff
@@ -0,0 +1,51 @@
+process_state_dictionary <- function(outfile, county_boundaries_census_sp) {
+  # get unique FIPS both from the sp data and dataRetrieval
+  unique_fips <- sort(unique(as.character(county_boundaries_census_sp@data$STATEFP)))
+  dr_states <- dataRetrieval::stateCd %>%
+    transmute(STATE_FIPS=STATE, STATE_ABBV=STUSAB, STATE_NAME=STATE_NAME)
+  
+  # check for unrecognized FIPS codes
+  bad_fips <- unique_fips[!(unique_fips %in% dr_states$STATE_FIPS)]
+  if(length(bad_fips) > 0) {
+    stop('these state FIPS are unknown to dataRetrieval: ', paste(bad_fips, collapse=', '))
```
lindsaycarr: It might be useful to state that we are using `dataRetrieval` as the authoritative source for county and state fips in the yaml file rather than buried down here.

scripts/preprocess/process_zip_to_shp.R
```diff
@@ -15,6 +15,10 @@ read_shp_zip <- function(zipfile) {
   # read the layer from the shapefile
   data_out <- rgdal::readOGR(shp_path, layer=layer, verbose=FALSE)
   
+  # filter out Northern Mariana Island (STATEFP=='69', ABBV='MP'), Guam
+  # (STATEFP=='66', ABBV='GU'), and American Samoa (STATEFP=='60', ABBV='AS')
```
lindsaycarr: are these something we will want to explicitly state on the viz somewhere?

lindsaycarr: like a footnote?

scripts/process/merge_centroids_wu.R
```diff
@@ -0,0 +1,14 @@
+process.merge_centroids_wu <- function(viz) {
+  deps <- readDepends(viz)
+  wu_data_15_simple <- deps[["wu_data_15_simple"]]
+  centroids_topo <- geojsonio::topojson_read(as.viz(viz$depends$centroids_topo)$location, stringsAsFactors=FALSE) # bypass readDepends for topojson
```
lindsaycarr: why do we need to bypass readDepends for topojson?

js/counties.js
aappling-usgs: It's not working on http://dev-owi.usgs.gov/vizlab/water-use-15/#view=SD&category=total right now, so probably not. I thought .text was supposed to be useful as a quick hovertext option, so I'm a little surprised to see that it wasn't doing anything (can you explain?)...but it was always just a placeholder anyway.

js/map_utils.js
aappling-usgs: Yes, `!proj` is `true` for `null` - it's not the most obvious syntax, but I've seen it in several examples, so maybe it's standard, and I am not sure what other options exist.

i like printing d because then i can see both the coordinates and the properties, including state/county/fips, which helped me diagnose. i just expand the value in the console.

scripts/preprocess/process_compute_centroids.R
aappling-usgs: yes, coming

scripts/preprocess/process_compute_centroids.R
aappling-usgs: `spdat` (aka `county_boundaries_census_sp`) and `county_dict_file` get merged in this file and in process_counties (with a few lines of code duplication in those two files, now noted in both files, that I don't think is worth effort to reduce further). Therefore the county boundaries, centroids, and state boundaries are all more final/authoritative than either spdat or county_dict_file.

In fact, i'll stop pushing the county and state dictionaries to S3 because we're getting away without using them directly in the viz.yaml.

examples of unnecessary ones are all of the columns from spdat@data not included in c(GEOID, STATE_ABBV, and COUNTY_LONG):
```r
> head(spdat@data)
  STATEFP COUNTYFP COUNTYNS       AFFGEOID GEOID    NAME LSAD      ALAND   AWATER
0      01      005 00161528 0500000US01005 01005 Barbour   06 2291820706 50864677
1      01      023 00161537 0500000US01023 01023 Choctaw   06 2365954971 19059247
2      01      035 00161543 0500000US01035 01035 Conecuh   06 2201896058  6643480
3      01      051 00161551 0500000US01051 01051  Elmore   06 1601876535 99850740
4      01      065 00161558 0500000US01065 01065    Hale   06 1667804583 32525874
5      01      109 00161581 0500000US01109 01109    Pike   06 1740741211  2336975
```

scripts/preprocess/process_dictionaries.R
aappling-usgs: i don't really think we are deferring to dataRetrieval for the FIPS codes; we're just checking for inconsistencies and making sure (by this check) that we'd know if there were any. we're just adding in STATE_ABBV and STATE_NAME using dataRetreival. i don't see that addition as rising to the level of needing a comment in the yaml

scripts/preprocess/process_zip_to_shp.R
aappling-usgs: i don't think so...we just don't have water use data for those territories. lots of vizzies get presented without justifying this omission (and even omit other places, e.g., PR and VI, without saying anything)

js/counties.js
lindsaycarr: I think that this only works if you've appended a text element, but not all elements allow text to be added. E.g. `.append("text").text("hi")` would work, but `.append("circle").text("hi")` does not.

scripts/process/merge_centroids_wu.R
aappling-usgs: we could add a readData function for topojson specifically, but we don't have one now, and I don't feel like it's worth it given that vizlab doesn't break right now and it's just one line of code to do it here. Still, I've added an issue to vizlab at https://github.com/USGS-VIZLAB/vizlab/issues/346, so maybe we'll have a readData.topojson function in the future.

js/counties.js
aappling-usgs: ah, that explains it =)

aappling-usgs:
ok @lindsaycarr , thanks for your thoughts. i've responded and pushed edits. ready to merge when you are.

Issue #24, opened 2018-02-26
Draft text about viz functionality


aappling-usgs:
See also #23, but we should draft the text on basics of how to use the vizzy.

Issue #63, opened 2018-03-02
Elegant style


aappling-usgs:
“if we go w/ a style, make it ridiculously consistent. So if we round a corner, we round all. And we do it the same (radius)” - jread

Can we lose the dark gray outline around the buttons? - lcarr

Do we want to mimic the behavior (zoom in/out, and click to adjacent states) from here: https://projects.fivethirtyeight.com/facebook-primary/ ?

lindsaycarr:
Should we circle outlines that are the slightly darker shade of that color? See the prototype figs for an example.

jread-usgs:
A couple of thoughts:
 - Mousing over any county circle should emphasize the backing state in the same way as mousing over the state under the circles. Clicking either a dot or state poly zooms you to the state view
 - We should consider showing the state borders at the national view scale
 - add a USGS watermark to the map
 - can we get an svg in for the header? it looks blurry on my comp compared to the sharpness of the rest of the viz elements
![image](https://user-images.githubusercontent.com/2349007/36938758-0dffbf20-1eec-11e8-9076-5fab7baadfb9.png)
 - the main title floating above the map: we'll want to figure out style for that, and I am thinking it is a placeholder for now. I also think it is kind of jarring that the title changes w/ the category switch
 - we might want to be consistent with cursor styles. For example, the category buttons flip to a "hand pointer" showing that you can click them, but other clickable things don't do the same. See OCD point above about ridiculous consistency
 - personal preference is to style svg and tooltip text as _not-highlightable_. I think it makes us look more pro when this can't happen:
![image](https://user-images.githubusercontent.com/2349007/36938800-a8b84a82-1eec-11e8-893e-20847d661069.png)
 - the footer. But that is captured in another issue.
 - I think the buttons need a little love design-wise

wdwatkins:
I second state borders on the national map.

I think we should definitely have the adjacent states and counties in some form on the zoomed in views, the dots look a bit odd just hanging in white space.

Also for the tool tips, I think the county names should include the state (e.g. `Dane County, WI`), and the population number needs a unit

lindsaycarr:
Tooltip values need help. MGD needs to be spelled out. And yes, population needs a unit - or just converted to "# of people" rather than "thousands of people" and rounded.

aappling-usgs:
For showing adjacent states and counties: agreed that what we have right now is not good, but we could go in either of two directions:

[...]

PR #76, opened 2018-03-14
A bit of text

ldecicco-USGS:


ldecicco-USGS:
image

Time ~15 minutes

ldecicco-USGS:
Fixes #24

data/description_text.yaml
```diff
@@ -18,19 +18,11 @@ description_text-markdown: |+
...
-  fermentum. Lacus laoreet non curabitur gravida arcu ac. Rutrum tellus 
-  pellentesque eu tincidunt tortor aliquam nulla. Massa eget egestas purus 
-  viverra accumsan in nisl nisi scelerisque. Faucibus turpis in eu mi 
-  bibendum neque egestas congue quisque. Purus sit amet luctus venenatis 
-  lectus magna fringilla urna. Odio ut enim blandit volutpat maecenas 
-  volutpat. Aliquam sem et tortor consequat id porta. Nunc lobortis mattis 
-  aliquam faucibus. Odio pellentesque diam volutpat commodo sed. Arcu dictum
-  varius duis at consectetur lorem donec massa sapien.
+  Explore how water is used throughout the United States. Click or 
+  hover over on any of the specific use-case boxes below the map to
```
lindsaycarr: I see where you're going with simplifying it, but it might be best to just say "category" or "Click or hover on the types of water use below the map"

data/description_text.yaml
```diff
@@ -18,19 +18,11 @@ description_text-markdown: |+
...
-  volutpat. Aliquam sem et tortor consequat id porta. Nunc lobortis mattis 
-  aliquam faucibus. Odio pellentesque diam volutpat commodo sed. Arcu dictum
-  varius duis at consectetur lorem donec massa sapien.
+  Explore how water is used throughout the United States. Click or 
+  hover over on any of the specific use-case boxes below the map to 
+  discover how thermoelectric, public supply, irrigation, or industrial 
+  water use is distributed. Click on a state to zoom in to more easily see the
+  difference in county use. Once zoomed in, simply click anywhere outside
+  the state on the map to return to the full country view. Hover (tap on 
+  mobile) over any of the individual counties to get more information.
```
lindsaycarr: Currently, you can only hover over the circle for each county, not the county itself.

aappling-usgs: True...but is that how we want it to be for all time? I've been wondering whether we should try to make the counties clickable as well.

lindsaycarr: Yes...but what if we don't end up doing it and forget to update the text?

jread-usgs: It would be great to be able to hover over the county that has the dot and get the tip for it. I assume this is a target for us, unless the county geoms are limiting performance in a major way. We also might want to get further down the path of what we want the end result to be before spending too much time altering the text. For example - does clicking a neighboring state once zoomed in pan you to that state-zoom, or zoom you out to national? I don't know that we've resolved some of these questions (or started the discussions).

aappling-usgs: Yeah, fair. Maybe there's no way to predict right now what the text will need to be. I'm ok with leaving the text as-is and remembering to refine the details once we've finalized the behavior.

Performance aside, is it going to be easy to get the hover behavior to be intuitive when it includes both counties and dots? big dots extend into other counties...when you hover over County A's dot where it extends into County B, which hover text should come up, and can we easily implement it that way?

I'm not sure how many of these design decisions need to turn into full-on discussions. There are many such decisions, and open-ended discussions are costly. I like the current zoom behavior, which is why I wrote it that way, though it's also absolutely possible to pan to a neighboring state. Jordan or others, if you have specific views and want to question the current behavior, could you create an issue proposing the approach you'd prefer? Then if anybody disagrees, it can be up to them to make a counter proposal within that issue, and if it becomes clear that the topic needs live discussion, we can plan to have that discussion during a team meeting.

jread-usgs: Gotcha. The question about click/zoom is in here: https://github.com/USGS-VIZLAB/water-use-15/issues/63 `Do we want to mimic the behavior (zoom in/out, and click to adjacent states) from here: https://projects.fivethirtyeight.com/facebook-primary/ ?`

aappling-usgs: oh, yep, there's a lot there

data/description_text.yaml
ldecicco-USGS: still want that changed?

PR #80, opened 2018-03-14
Legend + shared circle scaling

### LINKED ISSUE #66: Add legend for circle sizes
We had one of these for the original water use vizzy. We need one here, too.

I'm also wondering whether we could use the same scale for all categories - does that look OK? facilitates comparison if it's feasible.

lindsaycarr:
Fixes #66. It has taken 2 hrs so far.

Circles are scaled based on min and max of all category data.
Legend circles stay the same size but change color based on `activeCategory`.

image

image

### DIFF
--- a/scripts/process/find_range.R
+++ b/scripts/process/find_range.R
@@ -1,0 +1,18 @@
+#' Extract the min and max to use to scale circles in D3
+#' 
+process.find_range <- function(viz) {
+  
+  deps <- readDepends(viz)
+  wu_data_15_simple <- deps[["wu_data_15_simple"]]
+  col_names <- deps[["col_names"]][["new_names"]]
+  
+  water_use_df <- dplyr::select(wu_data_15_simple, col_names)
+  # wu_data_15_range <- data.frame(
+  #   min = min(water_use_df, na.rm = TRUE),
+  #   max = max(water_use_df, na.rm = TRUE),
+  #   stringsAsFactors = FALSE
+  # )
+  wu_data_15_range <- range(water_use_df, na.rm = TRUE)
+  jsonlite::write_json(wu_data_15_range, viz[["location"]])
+  
+}
--- a/scripts/process/find_range.R
+++ b/scripts/process/find_range.R
@@ -7,12 +7,8 @@ process.find_range <- function(viz) {
   col_names <- deps[["col_names"]][["new_names"]]
   
   water_use_df <- dplyr::select(wu_data_15_simple, col_names)
-  # wu_data_15_range <- data.frame(
-  #   min = min(water_use_df, na.rm = TRUE),
-  #   max = max(water_use_df, na.rm = TRUE),
-  #   stringsAsFactors = FALSE
-  # )
   wu_data_15_range <- range(water_use_df, na.rm = TRUE)
+  
   jsonlite::write_json(wu_data_15_range, viz[["location"]])
   
 }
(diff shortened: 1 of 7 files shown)


scripts/process/find_range.R
```diff
@@ -0,0 +1,18 @@
+#' Extract the min and max to use to scale circles in D3
+#' 
+process.find_range <- function(viz) {
+  
+  deps <- readDepends(viz)
+  wu_data_15_simple <- deps[["wu_data_15_simple"]]
+  col_names <- deps[["col_names"]][["new_names"]]
+  
+  water_use_df <- dplyr::select(wu_data_15_simple, col_names)
```
aappling-usgs: wooaaahh, hold up! you can pass a character vector in to `dplyr::select`? i guess you can!
```r
> df <- data.frame(x=1,y=2,z=3)
> dplyr::select(df, c('x','y'))
  x y
1 1 2
```
when did that happen?

scripts/process/find_range.R
```diff
@@ -0,0 +1,18 @@
...
+  # wu_data_15_range <- data.frame(
+  #   min = min(water_use_df, na.rm = TRUE),
+  #   max = max(water_use_df, na.rm = TRUE),
+  #   stringsAsFactors = FALSE
+  # )
```
aappling-usgs: looks like these lines could be removed now?

scripts/process/find_range.R
```diff
@@ -0,0 +1,18 @@
...
+  wu_data_15_range <- range(water_use_df, na.rm = TRUE)
+  jsonlite::write_json(wu_data_15_range, viz[["location"]])
```
aappling-usgs: not that we should do this right now, but these range data seem like a good use case for webpack or a templated app.js or something so we don't have to pass a bunch of small files around. i don't know what the performance cost of passing around 1 more tiny file is, but i bet it's bigger than having it embedded in another file to begin with.

viz.yaml
```diff
@@ -265,15 +281,18 @@ publish:
       hashes: url_hashes
```
aappling-usgs: would you mind renaming `social-section` to `social_section` and `national-water-use` to `national_water_use` while you're in here? the style police are really getting on my case about it =)

scripts/process/find_range.R
lindsaycarr: I don't know, I was surprised myself!

scripts/process/find_range.R
lindsaycarr: oops - yes. I was testing different formats

viz.yaml
lindsaycarr: on it!

scripts/process/find_range.R
lindsaycarr: I agree, I would love to be able to have a better connection between vizlab and js stuff in the future. I know that was one of my investigatory tasks in d3 before we got started on this, but I was never able to actually work on it.

scripts/process/find_range.R
aappling-usgs: well, it's cool! i will definitely be using this - thanks for teaching me things today

scripts/process/find_range.R
aappling-usgs: yeah, and i think that's fine. just wanted to keep collecting use cases so that when we do find the time and the right approach, we're clear on why we're doing it :+1:

aappling-usgs:
good to go now, right?

lindsaycarr:
yep!

PR #85, opened 2018-03-20
switch to sciencebase storage

wdwatkins:


aappling-usgs:
looks good. and you created those items and transferred the files?

aappling-usgs:
Assuming answer to above is Yes, i think we should merge this and tell our WiM collaborators that we're ready for them.

To finish issue #85 we'll also need to redirect the preprocessing posting steps to post to ScienceBase rather than S3. https://github.com/USGS-VIZLAB/water-use-15/blob/master/preprocess.yaml#L63-L70

wdwatkins:
Yes, it built successfully.  This does mean that you need to `authenticate_sb` before building now, should probably note that somewhere.

aappling-usgs:
Hmm, has that always been the case that you need to manually authenticate in vizlab, can't just use existing credentials? There's a line in there that ought to be doing the authentication for you once you have credentials set up: https://github.com/USGS-VIZLAB/vizlab/blob/master/R/fetch.R#L47

aappling-usgs:
I looked up the [code that does the authentication](https://github.com/USGS-VIZLAB/vizlab/blob/c06eb97ec1b670359236ed574ac4441ae23d5ecf/R/authRemote.R#L41). David's solution works; the full suite of options tried by authRemote.sciencebase is:

1) if user is already logged in, uses existing session, otherwise goes to 2
2) if dssecrets is installed, attempts to look up and use the credential from dssecrets, if that fails goes to 3
3) if secret is installed, attempts to look up and use the credential stored as `'vizlab-sciencebase'` with fields `username` and `password`, if that fails goes to 4
4) gives up

PR #87, opened 2018-03-23
Text has no margin

mwernimont:
This is how we added text in the Hurrican Marie Viz, https://github.com/USGS-VIZLAB/hurricane-maria/blob/master/viz.yaml so I copied that style. 

I believe this is how we are supposed to do it in the current version of Vizlab, as I don't feel great about messing with the printall publisher directly.  If you feel its a better decision let me know and we can discuss.

Vizlab will have a PR https://github.com/USGS-VIZLAB/vizlab/pull/362 adding the .viz-text class, which creates a basis for all the text going on into the future.

screen shot 2018-03-23 at 1 10 28 pm

### DIFF
--- a/layout/templates/textSection.mustache
+++ b/layout/templates/textSection.mustache
@@ -1,0 +1,4 @@
+<div class="viz-text">
+  {{{directions}}}
+  {{{description}}}
+</div>
--- a/viz.yaml
+++ b/viz.yaml
@@ -179,8 +179,9 @@ publish:
       custom_css: "custom_css"
       social: "social_section"
       socialCSS: "lib-social-media-css"
+      description_text: "description_text"
+      directions_text: "directions_text"
       header: "header_section"
-      text_section: "text_section"
       footer: "footer_section"
       hashes: "url_hashes"
       projection: "custom_projection"
@@ -197,7 +198,7 @@ publish:
       resources: ["lib_d3_js", "topojson", "header_css", "content_css", "footer_css", "socialCSS",
                   "map_css", "custom_css", "hashes", "projection", "map_utils", "map_counties", 
                   "legend_utils", "pie_utils"]
-      sections: ["social", "map", "text_section"] # adding header and footer here, doubled each on the page
+      sections: ["social","map","description_text", "directions_text"] # adding header and footer here, doubled each on the page
   -
     id: topojson
     location: "js/topojson.v1.min.js"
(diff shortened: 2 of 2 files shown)


mwernimont:
Also was thinking about creating a Vizlab template to just add text to each time, but for the text to show up in the sections you have to add the context {{{direction}}} directly into the template.  Which is hard to do if the template is living in Vizlab and not the current Viz repo.  Hopefully this makes sense.

layout/templates/textSection.mustache
```diff
@@ -0,0 +1,4 @@
+<div class="viz-text">
```
ldecicco-USGS: Can we put this <div class="viz-text"> in the printAll template in vizlab? That way, we don't need to make a textSection mustache for each viz. We still could...but we wouldn't have to.

mwernimont: No clue, can we?

aappling-usgs: yes, i think laura has it. modify this: https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/templates/printall.mustache. also helps with future vizzies already having divs with `viz-text` class by default so that the styling added in https://github.com/USGS-VIZLAB/vizlab/pull/362 will get applied

mwernimont:
Looks to be working on my local version, so should be good to go.

viz.yaml
```diff
@@ -302,4 +295,18 @@ publish:
       resources: ["lib_d3_js", "topojson", 
                   "map_css", "custom_css", "hashes", "projection", "map_utils", 
                   "map_counties", "legend_utils", "pie_utils"]
-      embed: ["map", "text_section"]
+      embed: ["map","directions_text"]
+  -
+    id: description_text
```
aappling-usgs: ah, now i see the beauty of your first approach of defining a new template - i did like the consolidation into one text_section item. but even so i think this most recent change/reversion is still the way to go to support future vizzies.

ldecicco-USGS: And we can still do that approach, right?

aappling-usgs:
@ldecicco-USGS , both this one and https://github.com/USGS-VIZLAB/vizlab/pull/362 are looking good to me, so i'm ready to merge, but you've been more closely involved in this area recently - do you approve the current configuration?

viz.yaml
aappling-usgs: oh! i guess i don't understand the system all that well, because i thought we must need `context : text` to be named just like that to work with the `printall` template. keeping consolidated would be excellent - how would it work?

ldecicco-USGS: Maybe I'm not really following this conversation correctly. I think anytime you can define your own template in an individual vizlab project. And now as a bonus, if you wrap it in `<div class="viz-text">`, you'll get Marty's fancy style. But, if you just want to push a little text in like we've done it historically in water-use (using the vizlab printAll template), then you still get the nice style. Anywho, I approve this and the other pull request.

ldecicco-USGS:
We'll need to update the Jenkin's job to get the newer vizlab

PR #101, opened 2018-03-28
plump PR

mwernimont:
Made PR and VI 3 times bigger and moved them north.
screen shot 2018-03-28 at 10 47 40 am

### DIFF
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -78,7 +78,7 @@ function albersUsaTerritories() {
 
     return (y >= 0.120 && y < 0.234 && x >= -0.425 && x < -0.214 ? alaska
         : y >= 0.166 && y < 0.234 && x >= -0.214 && x < -0.115 ? hawaii
-        : y >= 0.2064 && y < 0.2413 && x >= 0.312 && x < 0.385 ? puertoRico
+        : y >= 0.130 && y < 0.190 && x >= 0.280 && x < 0.410 ? puertoRico
         : lower48).invert(coordinates);
 
   };
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -101,7 +101,7 @@ function albersUsaTerritories() {
     lower48.scale(_);
     alaska.scale(_ * 0.35);
     hawaii.scale(_);
-    puertoRico.scale(_);
+    puertoRico.scale(_ * 3);
     return albersUsa.translate(lower48.translate());
   };
 
@@ -176,8 +176,8 @@ function albersUsaTerritories() {
         .stream(pointStream);
 
     puertoRicoPoint = puertoRico
-        .translate([x + 0.350 * k, y + 0.224 * k])
-        .clipExtent([[x + 0.312 * k + epsilon, y + 0.2064 * k + epsilon],[x + 0.385 * k - epsilon, y + 0.233 * k - epsilon]])
+        .translate([x + 0.340 * k, y + 0.170 * k])
+        .clipExtent([[x + 0.280 * k + epsilon, y + 0.130 * k + epsilon],[x + 0.410 * k - epsilon, y + 0.190 * k - epsilon]])
         .stream(pointStream);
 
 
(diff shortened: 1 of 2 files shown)


js/custom_projection.js
```diff
@@ -176,8 +176,8 @@ function albersUsaTerritories() {
         .stream(pointStream);
 
     puertoRicoPoint = puertoRico
-        .translate([x + 0.350 * k, y + 0.224 * k])
-        .clipExtent([[x + 0.312 * k + epsilon, y + 0.2064 * k + epsilon],[x + 0.385 * k - epsilon, y + 0.233 * k - epsilon]])
+        .translate([x + 0.340 * k, y + 0.170 * k])
+        .clipExtent([[x + 0.280 * k + epsilon, y + 0.130 * k + epsilon],[x + 0.410 * k - epsilon, y + 0.190 * k - epsilon]])
```
aappling-usgs: cool, nice work figuring this out! do you think https://github.com/USGS-VIZLAB/water-use-15/pull/101/files#diff-1b045ffa0e22d4eccda9408d9c5d9777R81 might also need to be updated to match these changes?

mwernimont: So I'm confused on this the link just leads me to what I just changed.  Did you mean to do that?

mwernimont: Never mind, needed to scroll up

mwernimont: I wouldn't change those, I'm not exactly sure but I feel those numbers have a higher purpose of how PR is projected, and the numbers I'm changing are just where PR is projected and how much of it is projected.  

The problem right now there is some uncertainty about what all these numbers do for the projection.  I only honestly feel comfortable making changes to the ones I did.

May not be an answer you find suitable, but its what I got at the moment.  Definitely, more study needed in projections.

aappling-usgs: I believe the numbers I highlighted would be used if you wanted to be able to capture a set of svg coordinates (e.g., from a user's click) and turn that back into latitude and longitude (https://github.com/d3/d3-3.x-api-reference/blob/master/Geo-Projections.md#invert). That functionality is not important to water-use-15 functionality today but could be important to this or another viz in the future.

Playing around in the console:
```
// i think this is on alaska in the svg, goes to alaska in lon/lat
projection.invert([150,550])
(2) [-150.63359833273702, 64.08072856727983]
// https://www.google.com/maps/place/60%C2%B055'41.7%22N+149%C2%B007'00.5%22W/@63.5931454,-159.7661388,4.6z/data=!4m5!3m4!1s0x0:0x0!8m2!3d60.9282557!4d-149.1168018?hl=en

// just changing the y coordinate, but note that the x (and y) coords change a lot, so we're not in AK anymore
projection.invert([150,450])
(2) [-116.53973890285776, 32.118858217819]
```

Having looked a little more, I'm feeling increasingly confident that those numbers i highlighted should be adjusted to match your new numbers down here; also see how both alaska and hawaii match up almost perfectly? So I'm pushing a little harder - I advocate for changing those numbers now so they don't trip us up in the future, if and when we need to use `invert`.

mwernimont: Alrighty, I'll trust you

aappling-usgs: Trust now, assign blame later, whatever you need to do =)

layout/templates/textSection.mustache
```diff
@@ -1,4 +0,0 @@
-<div class="viz-text">
```
lindsaycarr: confused..is this just extra garbage you are cleaning up or did it get mixed up in this PR accidentally?

mwernimont: That is a cleanup, we aren't using a template for the text, so not needed.

mwernimont: I shoulda explained that.

lindsaycarr: No worries - just wanted to make sure it wasn't swept up in a merge conflict resolution or something!

lindsaycarr:
Was going to merge, but I think there was one more change this is waiting on, right? Adjusting those numbers @aappling-usgs pointed out?

mwernimont:
Yep, I need to do that quick.

mwernimont:
@aappling-usgs nothing seemed to change visually, so I think I should just trust you more often.

Issue #97, opened 2018-03-27
Improve pie chart / circle performance


aappling-usgs:
Make it faster to load the page, and transition into and out of pie charts and circles, 

Relates to #71 (basic pie charts), #72 (spiffy pie chart transitions), and #62 (load time performance), and messing with transitions and loadable data are within scope (despite overlapping with those other issues) to the extent that they affect performance.

aappling-usgs:
Lindsay and Alison have already been working on this for several hours today (https://github.com/lindsaycarr/water-use-15/commit/eafe0893aa0c10940ded20e90ad1dc54bfedf489 through https://github.com/lindsaycarr/water-use-15/commit/0269bf314869fabf7e1d3a07213ea4adf8b1c188). We've made it so the pie groups and pie slices only get created once per session, moved the pie chart data preparation into build_map so that also only happens once now, and temporarily turned off transitions to be able to view the changes faster. As a baseline from there (I'm hoping to find additional opportunities for improvement), here are 
* load times to `view=USA&category=total` (milliseconds):

total | scripts (all the yellow) | rendering (all the purple) | painting (all the green) | 'load' handler took __ms
-- | -- | -- | -- | -- 
1507 | 1229 | 77 | 5.5 | 873
2745| 1351 | 82 | 5.3 | 949
1625 | 1313 | 79 | 5.3 | 940
1551 | 1246 | 79 | 5.6 | 873
1556 | 1254 | 80 | 5.3 | 900

* load times to `view=USA&category=piechart` (milliseconds) (with `Error: <circle> attribute r: Expected length, "NaN".`):

total | scripts (yellow) | rendering (purple) | painting (green) | 'load' handler
-- | -- | -- | -- | -- 
2967 | 1749 | 80 | 3.4 | 1201
2166 | 1782 | 77 | 3.2 | 1372
3204 | 1776 | 75 | 3.3 | 1390
2096 | 1728 | 81 | 3.9 | 1340
2137 | 1776 | 85 | 2.5 | 1404

* times for first transition from Total to Pie view

total | scripts (yellow) | rendering (purple) | painting (green) | 'mouseover' handler
-- | -- | -- | -- | -- 
?? | 473 | 302 | 26.6 | 428
?? | 520 | 310 | 27.5 | 464
?? | 536 | 307 | 28.4 | 497
?? | 517 | 303 | 27.5 | 474
?? | 561 | 301 | 26.7 | 534

[...]

PR #102, opened 2018-03-28
Piechart performance

### LINKED ISSUE #82: Pie chart performance
Nodes keep appending, rather than update (for pies).

image

After hovering over pie chart and then back:

image

lindsaycarr:
See #82 and #97. I think this closes #82.


js/pie_utils.js
```diff
@@ -1,51 +1,58 @@
...
+  piesBaked = true;
+  
+  // these newly added pie charts won't work until updatePieCharts is called, but
+  // since this function always gets called from updatePieCharts, that should be fine
+}
+
+function updatePieCharts() {
+  
+  // add the pies if needed
+  if (!piesBaked) {
```
jread-usgs: 😂

aappling-usgs:
I don't think I can do a good job of reviewing my own code, but Lindsay and I pair programmed the first half of this PR, and Lindsay reviewed my work on the second half already, so I'm going to declare this sufficiently reviewed for merging.

PR #103, opened 2018-03-28
expand clipping region to include one little county-was causing proje…

lindsaycarr:
…ction error

noticed that changes in #101 cut one little county off in PR. Just expanded so that we don't get a projection error and can keep on moving!


aappling-usgs:
Could you also edit the other line that I made Marty edit? Line 81: https://github.com/USGS-VIZLAB/water-use-15/pull/103/files#diff-1b045ffa0e22d4eccda9408d9c5d9777R81

lindsaycarr:
oops - completely meant to do that

Issue #96, opened 2018-03-26
Darken the circle outlines


aappling-usgs:

jread-usgs:
A starting point for this would be to use 100% opacity of the full color for the stroke and use a lesser opacity for the fill. That is what I did for the gif and I think it looks ok.

jread-usgs:
One thing that makes this a little harder than it would seem is that the stroke width will scale with the circle, so you need to scale _down_ the stroke when the circle scales up to keep the same line width. 

Same thing goes for state borders. Easy way to see this is in Rhode Island, where we'd want the circle borders to be the same as national view and be able to limit the state border scale in the same way:
![image](https://user-images.githubusercontent.com/2349007/38063473-fdfe8976-32be-11e8-92d9-43e2236c6d8f.png)

***Actually*** this seems to be only an issue w/ zooming into the states. D3 scale seems to be smart and handle the stroke scaling itself

aappling-usgs:
Revised the title to describe the expanded scope from jread's comments

I think there are two ways to preserve the same line widths at both national and state scales:
1) use changes to projection rather than changes to transform to do the zooming. i'm not quite sure how to do this, but it might be faster. or not?
2) scale down all the stroke widths in proportion to the scaling up from the transform. zooming happens in `updateView` [here](https://github.com/USGS-VIZLAB/water-use-15/blob/e6072cbd4e0251eefd38dd561c32351ef0254d74/js/map_utils.js#L119-L148) and [here](https://github.com/USGS-VIZLAB/water-use-15/blob/e6072cbd4e0251eefd38dd561c32351ef0254d74/js/map_utils.js#L173-L179). This approach might be faster to implement, and who knows - maybe it's actually the better performance choice, too.

jread-usgs:
Note that two options help uw w/ stroke width on the circles: use radius scaling instead of `transform=scale(x,y)` or use `path` instead of `circle` to represent the dots

aappling-usgs:
i think what's left in this issue is to switch from using transforms to using radius scaling. is that right? anything else? path instead of circle is in issue #119...does this issue (#96) need to wait on that one (#119) to proceed?

jread-usgs:
If we are using paths in the end, should we skip radius scaling?

jread-usgs:
see https://github.com/USGS-VIZLAB/water-use-15/issues/97#issuecomment-377971764 for the details, but here is an example:
```xml
[...]

Issue #72, opened 2018-03-12
Spruce up pie chart transitions


aappling-usgs:
Follows on #71

lindsaycarr:
Might be interesting to make all of views pie charts but filter data so that it's only one category. Then I think the fancy windshield wiper transitions would work.

https://bl.ocks.org/rshaker/225c6df494811f46f6ea53eba63da817

jread-usgs:
We've discussed a few transition options that might help w/ performance. 

The _windshield wiper_ looks good, but if it isn't smooth and dogs the browser it won't look so good
The _color transition_ seems to be a simpler option, but maybe the color interpolation is going to be challenging to make smooth
Another idea is the _coin flip_ transition, which takes advantage of being able to avoid the color interpolation **and** letting us stagger the transitions left to right (if that helps us) while making it look purposeful. 

for the _coin flip_ transition, say we are doubling the size:
If the transition total time is 1 second, the first 0.5s scales x from 1 to 0 while scaling y from 1 to 1.5. At time 0.5s, you are looking at the "edge of the coin" (only the stroke width) and we instantly flip the fill color(s) and add/remove pie paths (if relevant). Then from 0.5 to 1s we scale x from 0 to 2 and y from 1.5 to 2.

It seems the color transition may be one of the more expensive things happening in our various options for transitions [source](https://www.html5rocks.com/en/tutorials/speed/high-performance-animations/) but note that reference is pretty dated.

aappling-usgs:
We're trying lots of things, but at this point I'm finding it hard to separate this issue conceptually from #97, so i'm going to close this one in favor of just adding comments to that issue.

PR #105, opened 2018-03-29
Coin flips between 

aappling-usgs:
@lindsaycarr I'm not going to get to a complete PR today, so making this PR into a new branch (`coin-flip`). It needs work. Feel free to try playing with and debugging it, or I'll get to it once I'm in the office.

I started out attempting to just merge the circle and pie slice svg elements in some way, started thinking about all the various transitions that would be available/appealing, and then switched directions when I saw Jordan's suggestion of using coin flip transitions, which ended up affecting a lot of decisions about how I did the rewrite.

One thing I'm trying is thinking of pies as groups containing both slices (as before) and a tin (the circles). This allows us to do all transforms and scaling at the group level, for both slices and tins, which should be really useful for coin flip transitions.

So far the coin flips are looking quite choppy, and there are also still errors popping up because it was too big a rewrite for me to get correct right away. I think it's doable, but there's more debugging to do for sure.

I'm going to get in trouble if I keep coding tonight, so there you have it. Happy to discuss/explain more tomorrow.

### DIFF
--- a/js/pie_utils.js
+++ b/js/pie_utils.js
@@ -9,14 +9,14 @@ function addPieSlices() {
   //relies on map as a global variable
   
   // add pie slices to each pie group
-  map.selectAll('g.pie').selectAll('.pieslice')  
+  map.selectAll('g.pie').selectAll('.slice')  
     .data(function(d) {
       return d.sliceGeomData;
     })
     .enter()
     .append("path")
-      .classed("pieslice", true)
-      .attr("fill", function(d) { 
+      .classed("slice", true)
+      .style("fill", function(d) { 
         return categoryToColor(d.data.category); 
       })
       .attr("d", arcpath);
@@ -27,16 +27,21 @@ function addPieSlices() {
   // since this function always gets called from updatePieCharts, that should be fine
 }
 
-function updatePieSlices(prevTransition) {
+function updatePieSlices(category, delay, duration) {
   
   // add the pies if needed
   if (!piesBaked) {
     addPieSlices();
   }
   
-  var newTransition = map.selectAll('.pieslice') 
-    .transition(prevTransition).duration(0);
-    
-  return newTransition;
+  if(category === 'piechart') {
+    d3.selectAll(".slice")
+      .transition().delay(delay).duration(duration)
+      .style("display", null);
+  } else {
+    d3.selectAll(".slice")
+      .transition().delay(delay).duration(duration)
+      .style("display", "none");
+  }
 }
 
--- a/js/pie_utils.js
+++ b/js/pie_utils.js
@@ -1,36 +1,20 @@
 // functions related to the pie charts
 
-var pie = d3.pie()
-  .sort(null)
-  .value(function(d) { return d.value; });
-
-var arcpath = d3.arc()
-  .innerRadius(0);
-
-function addPieCharts() {
+function addPieSlices() {
   
-  //relies on map and pieFormData as a global variable
-  
-  // each pie is a group of pie slices. here add the pie groups
-  var pies = map.selectAll('.pie')
-    .data(pieFormData);
-  var piesMerged = pies
-    .enter()
-    .append('g')
-      .classed("pie", true)
-      .attr("transform", function(d) {
-        return "translate(" + d.coordinates.x + "," + d.coordinates.y + ")";
-      })
-      .merge(pies);
+  //relies on map as a global variable
   
   // add pie slices to each pie group
-  piesMerged.selectAll('.pieslice')  
+  map.selectAll('g.pie').selectAll('.pieslice')  
     .data(function(d) {
-      return d.sliceGeomData;
+      return [d.sliceGeomData];
     })
     .enter()
     .append("path")
-      .classed("pieslice", true);
+      .classed("pieslice", true)
+      .attr("fill", function(d) { 
+        return categoryToColor(d.data.category); 
+      });
   
   piesBaked = true;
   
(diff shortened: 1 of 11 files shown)


aappling-usgs:
Next steps:
1. if switching between pie slices and tins, hide the other polygons
2. there's some attribute transform error coming up

jread-usgs:
The tin/pie analogy is also useful for when we apply a stroke for #96 . Using a stroke on the circles is simple, but the pie is hard because you don't actually want the radials to have strokes, just the crusts. But that's hard because there isn't a clear way to do that with a path that is easy. If you use tins and they are under the pies, in pie-mode your tin can have a greyish stroke under the semitransparent pies, (but no fill) and it will look like the crusts are 👍

js/symbol_utils.js
```diff
@@ -0,0 +1,112 @@
...
+  
+  // phase 1: sort and then transition width to 0, height to halfway to new value
+  t1 = pies
+    .sort(function(a,b) { 
+      return d3.descending(a.properties[[category]], b.properties[[category]]);
+    })
+    .transition().duration(500)
+    .attr("transform", function(d) {
+      var halfheight = (scaleCircles(d.properties[[oldCat]]) + scaleCircles(d.properties[[newCat]])) / 2;
+      return "translate(" + d.coordinates.x + "," + d.coordinates.y + ")"+
```
jread-usgs: if I read this correctly, this `translate` is being applied to each pie/tin group twice for each transition. This is what gets the dot in the right position. I think this is necessary to do each time because some pie/tins are added or removed at each category transition, but I am not sure. 

If we keep the same number of total county pie/tin groups (i.e, we don't drop the ones that are zero in the category), can we avoid running this translate each time and only use it once (when setting up the page)? I know the `scale` is necessary each time, but would reducing/removing the `translate()` calls during transitions help us with performance?

jread-usgs: Another thought. We ended up switching from js transitions to css transitions for water-use ( [code here](https://github.com/USGS-VIZLAB/water-use/blob/e3f0c454787f69bd56184ab55cd01dafd941acea/layout/js/map.js) ). We used scale3d() after walker looked into different options. Not sure if this is moot now that we are using d3, but css transforms may be a quick option to try (using style instead of transform attrs).

aappling-usgs: We're not currently adding or removing pie elements ever. My idea (not yet implemented) was to hide the unused ones (`display: false`?) but keep everything in place once it's created. I don't actually know whether that's faster or slower than fully removing and re-adding elements at each transition, but my suspicion is it's faster to keep them in place. That's another thing we could experiment with.

Interesting thought to separate the `translate` from the `scale` and only apply the `translate` bit once. I was imagining that any `translate` attribute led to a recalculation each time something was repainted anyway, but again, I suppose I don't actually know.

Notes to myself / other noobs regarding css transitions, which I've never used:
* in the linked code, these lines https://github.com/USGS-VIZLAB/water-use/blob/e3f0c454787f69bd56184ab55cd01dafd941acea/layout/js/map.js#L82-L87 and this line https://github.com/USGS-VIZLAB/water-use/blob/e3f0c454787f69bd56184ab55cd01dafd941acea/layout/js/map.js#L97 seem important.
* here's a concise comparison of d3 vs css transitions: https://stackoverflow.com/questions/46675499/animating-using-d3-transitions-vs-using-css-animation.
* you can chain css transitions: https://css-tricks.com/using-multi-step-animations-transitions/

The existing d3 transitions in this PR definitely need help, not just for speed but because they're not yet doing things in the right order. I think the middle phase is failing to take its place in a chain of transitions and is instead happening ASAP. we could try delays, or try to learn more about how to chain transitions at one level (g.pie) with transitions at a lower level (.pieslice, .tin) and back to transitions at the higher level (g.pie).

aappling-usgs:
Cross-referencing to Jordan's description of the coin flips approach: https://github.com/USGS-VIZLAB/water-use-15/issues/72#issuecomment-377092558

js/symbol_utils.js
jread-usgs: Cool - yes, those were the lines I was thinking of (was on mobile when I posted and it doesn't seem to let you link to line numbers when in mobile). 

I think you want `display: none;` for getting rid of pies (marty?)

Another advantage _I think_ CSS transitions give us is their handling of being interrupted. Again, I don't know if D3 is smarter than the way we used to use js to animate transitions, so maybe that point is moot. I am referring to the case where you mouse over one category and then another before the transition is complete.

aappling-usgs: D3 is smart about transition interruptions in the sense that if the user initiates a new transition, the old one gets abandoned. I suspect that's the desired behavior, but we should look at that specifically to make sure we're getting what we want with whatever implementation we land on.

js/symbol_utils.js
```diff
@@ -0,0 +1,112 @@
+// functions related to the symbol groups
+
+// each pie is a group of pie slices and a pie tin (a circle). here add the pie groups
```
lindsaycarr: so this wouldn't really be decreasing the number of nodes from the current implementation, right? There would still be 3223 circles + 3223 pie groups (with 5 slices each).

js/symbol_utils.js
lindsaycarr: I did not know you could bake transitions directly into CSS. The reason we've been applying fill styles via D3 is so that we can fade colors in and out, versus changing the class which you can't slow down. Good trick to be able to stick transitions directly into CSS land.

lindsaycarr: Based on that Stackoverflow comment from Alison's link, it looks like simple transitions in CSS will improve performance, but more complex ones are hard to create in just CSS so you might need to stick to D3.

aappling-usgs: true. however, coin flips ARE pretty simple transitions. i think it's probably possible, unfamiliar though it is to the two of us.

js/symbol_utils.js
aappling-usgs: yeah. i spent a long time drawing on paper to come up with a good plan for transitions, but things got complicated when i started thinking about transitions from one single category to another, and from pies to total, and from single categories to total - those things don't lend themselves as well to the wiper transitions, and so i more or less came to terms with needing a circle in the mix anyway. also, when i tried some wiper transitions they were cripplingly slow.

js/symbol_utils.js
lindsaycarr: Right - we should most certainly try out transitioning via CSS for coin flips! I wasn't trying to imply this one was hard, but was confirming the use cases for d3 to transition vs css.

aappling-usgs: oh, yes, agreed! for example, i doubt css could do wiper transitions, which involve recalculating the contents of the `d` path element string. (then again, i still know next to nothing about css transitions, so...)

ps the wiper transitions were looking pretty cool except for being slow. we should totally try them in some future viz.

js/pie_utils.js
```diff
@@ -1,101 +1,47 @@
...
-      return scaleCircles(d.data.total);
-    }))
-    .attr("fill", function(d) { 
-      return categoryToColor(d.data.category); 
-    });
-  
+  if(category === 'piechart') {
+    d3.selectAll(".slice")
+      .transition().delay(delay).duration(duration)
+      .style("display", null);
```
lindsaycarr: I think you might want `block` here?

aappling-usgs: can you tell me why? i don't know much about the "display" options, but i think when i first create the item there is no `display` info inline at all, so i just set it back to not being in there by assigning `null`

aappling-usgs: cool, thanks for the link!

PR #108, opened 2018-03-30
Circle as slice

aappling-usgs:
(oops, #107 was trying to pull into `master` instead of `coin-flip`)

Uses a pair of closely nested groups to control transform and scale separately.

```
- g.pie-locations
  - g.pie-scales
    - circle.tin
    - path.slice
    - path.slice
    - path.slice
    - path.slice
    - path.slice
```

Jordan says, "I think it is faster, but it still chokes on the pie transition"


aappling-usgs:
Given that Jordan wrote this and I reviewed it already, I'm not going to request another review.

PR #109, opened 2018-03-30
Coin flip

aappling-usgs:
I think this branch contains the way forward, so I think we should merge into master.

We may need to undo or change some of the specifics as we further explore our options for performance, but this runs correctly, has clearer separation of functions into files, and is structured to allow us to pretty easily try out some of the more promising options for handling transitions (playing with transition types and duration, seeking more ways to do required things just once, etc.).


lindsaycarr:
Woo! Does this touch any issues other than #97?

aappling-usgs:
I believe it's only #97, insofar as that one subsumes #72 and #82.

Issue #98, opened 2018-03-28
Create color palette mock-ups (S)


jread-usgs:
***I think we should wait on this task*** until we know if we are using the pie charts: 
A couple of notes here - 

Colorblind friendly becomes more important if we use the pie charts. Otherwise, we aren't doing a side-by-side of any categories, so it is more important to understand which category is _on_ than to see the color differences between categories. 

We'll want a blend between colors that make sense for the categories (e.g., green), aesthetically pleasing combinations, and ability for colorblind to pick out the differences.

jread-usgs:
We have decided to remove the pie charts from the interactive viz

aappling-usgs:
But note that we're still talking about including a static version of the pie charts in a key story

lindsaycarr:
Are colors for background map, states at national view, state color in zoom, boundaries, etc are being chosen? Just wondering if it's my job to do that with #95 or not.

jread-usgs:
I think this is done, but we should create another issue for doing a comprehensive style unification sweep a few days before our next milestone. Revisiting the state borders and zoom styling would be included in that. 
I don't think it is worth getting uber picky about right now since we are still iterating on the look and feel of certain things, including mobile and the category selectors in #89

aappling-usgs:
the colors Jordan has proposed are these:
```
var stateStyle = {
  nationView: {
    active: {
      'fill': 'white',
      'stroke': 'black',
      'stroke-width': 1.5
    },
    inactive: {
      'fill': 'white',
      'stroke': 'black',
      'stroke-width': 1.0
    }
  },
  stateView: {
    active: {
      'fill': '#DCDCDC',
      'stroke': 'black',
      'stroke-width': 1.5
    },
    inactive: {
      'fill': 'transparent',
      'stroke': 'black',
      'stroke-width': 1.0
    }
  }
};
+  if (category == "total") { return "#2678b2"; }
 +  else if (category == "thermoelectric") { return "#edc948"; }
 +  else if (category == "publicsupply") { return "#76b7b2"; }
 +  else if (category == "irrigation") { return "#59a14f"; }
 +  else if (category == "industrial") { return "#e15759"; }
 +  else if (category == "other") { return "#A9A9A9"; }
```

Which gives these colors:
![image](https://user-images.githubusercontent.com/12039957/38375672-99b31390-38ab-11e8-84d7-1cf6007b2e0f.png)

Which, since they're in the vizzy, are not just prototyped but are already in action.

[...]

PR #112, opened 2018-03-31
for #96 and some of #98 (trying a plain white look)

jread-usgs:
This isn't really complete because the circles scale (even w/o zoom) and have different strokes to start with. Tracking and correcting that may be a whole different can of worms. What I have here is some style work and the state border zoom factor correction on the stroke-width:

image

### DIFF
--- a/js/styles.js
+++ b/js/styles.js
@@ -2,26 +2,26 @@
 var stateStyle = {
   nationView: {
     active: {
-      'fill': '#BEBEBE',
-      'stroke': 'transparent', // looks OK white, too
-      'stroke-width': 0
+      'fill': 'white',
+      'stroke': 'black', // looks OK white, too
+      'stroke-width': 1.5
     },
     inactive: {
-      'fill': '#DCDCDC',
-      'stroke': 'transparent', // i think we're avoiding borders usually?
-      'stroke-width': 0
+      'fill': 'white',
+      'stroke': 'black', // i think we're avoiding borders usually?
+      'stroke-width': 1.0
     }
   },
   stateView: {
     active: {
       'fill': '#DCDCDC',
-      'stroke': 'transparent', // no need for border when there's fill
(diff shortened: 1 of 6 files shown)


js/styles.js
```diff
@@ -2,26 +2,26 @@
...
+      'stroke': 'black', // i think we're avoiding borders usually?
+      'stroke-width': 1.0
     }
   },
   stateView: {
     active: {
       'fill': '#DCDCDC',
-      'stroke': 'transparent', // no need for border when there's fill
-      'stroke-width': 0
+      'stroke': 'black', // no need for border when there's fill
```
aappling-usgs: could you take out the comments about avoiding borders now that we're not avoiding them?

js/symbol_utils.js
```diff
@@ -43,6 +43,7 @@ function updatePies(category, prevCategory) {
         return "scale(0.1 " + halfheight + ")";
       })
     .transition()
+      .ease(d3.easeLinear)
```
aappling-usgs: oh, good idea

js/styles.js
aappling-usgs: also, looks from your screenshot like state borders are coming through as grey despite being told to be black here. do you think that could be because county borders are plotted on top of state borders?

jread-usgs: Yeah, there are a couple of funky things going on w county borders and I was going to keep it on the back burner to see what we are going to do about z index. One thing is I think the counties appear over the state border, so they cover up the black border. Other issue I can into is if you load the page on a state zoom, the counties don’t come in until you zoom out to USA and then back in. I didn’t want to get too in the weeds on those issues since I didn’t know if there would be code changes related to the mouse overs (and related to load times waiting on the county data to come in).

jread-usgs: Excuse my typos and bad grammar. Am on mobile at the airport :)

PR #116, opened 2018-04-02
reinstate county hover and click, not circles

### LINKED ISSUE #92: National view: give circles same hover & click behavior as their backing state (S)
* Mousing over any circle should emphasize the backing state in the same way as mousing over the state under the circles (for circles that extend beyond their state boundary, does "backing state" mean the one directly behind the mouse or the one represented by the circle?) (S)
* Clicking either a dot or state poly zooms you to the state view (S)

lindsaycarr:
Fixes #92. Tooks <5min.

### DIFF
--- a/layout/css/map.css
+++ b/layout/css/map.css
@@ -24,8 +24,8 @@
   stroke-width: 0;
 }
 
-.county-point,
-.pieslice {
+.tin,
+.slice {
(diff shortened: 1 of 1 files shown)


aappling-usgs:
Oh, oops, that was my bad. Thanks for fixing.

PR #118, opened 2018-04-02
Outline styles

jread-usgs:


scripts/sandbox/generate_circle_path_test.R
```diff
@@ -0,0 +1,151 @@
...
+      stroke:red;
+      stroke-width:0.3;
+			fill-opacity:0.8;
+		}")
+svg <- bd %>% 
+  xml_add_child('svg', xmlns="http://www.w3.org/2000/svg", 
+                "xmlns:xlink"="http://www.w3.org/1999/xlink", version="1.1", preserveAspectRatio="xMinYMin meet", 
+                viewBox=sprintf("0 0 %s %s", view_coords[1], view_coords[2]))
+
+path_template <- "M%1.1f %1.1f a %1.1f %1.1f 0 1 1 0 0.01 " #x start, y-start, radius, radius
```
jread-usgs: This is how the `d` values for a path are generated. 

The pattern is creating json w/ an element for each category. But, there is a ton of redundant information and we could collapse this down if we thought the json was too big - we really just need x,y,radius and could build the json in the browser if needed. But I think that gzipping would take care of those redundant pieces of info anyhow so this wouldn't be necessary.

PR #122, opened 2018-04-02
Watermark

ldecicco-USGS:


js/watermark.js
```diff
@@ -0,0 +1,54 @@
...
+  var xPosWatermark = chart_width*0.01,
+      yPosWatermark = 0.95*chart_height;
+  
+  // add group for legend elements & move into the Atlantic Ocean
+var watermark = svg
+      .append('g')
+        .attr('transform', 
+              'translate('+xPosWatermark+','+yPosWatermark+')scale(0.25)')
+        .append('a')
+        .attr("xlink:href", "https://www2.usgs.gov/water")
```
lindsaycarr: do we not want to link to the new drupal site? https://www.usgs.gov/science/mission-areas/water-resources?qt-mission_areas_l2_landing_page_ta=0#qt-mission_areas_l2_landing_page_ta

js/watermark.js
```diff
@@ -0,0 +1,54 @@
...
...3.571-3.04-3.927-3.313-11.204-7.867-21.646-5.476-26.149-3.802-1.362 0.544-2.665 1.287-3.586 1.869l-28.602 19.13v34.666h116.03v-24.95c-2.55 1.62-18.27 10.12-40.063-10.46m-44.677-42.322c-0.619-0.578-1.304-1.194-1.915-1.698-13.702-10.6-26.646-5.409-29.376-4.116v11.931l6.714-4.523s10.346-7.674 26.446 0.195l-1.869-1.789m16.028 15.409c-0.603-0.534-1.214-1.083-1.823-1.664-12.157-10.285-23.908-7.67-28.781-5.864-1.382 0.554-2.7 1.303-3.629 1.887l-13.086 8.754v12.288l21.888-14.748s10.228-7.589 26.166 0.054l-0.735-0.707m68.722 12.865c-4.563 3.078-9.203 6.203-11.048 7.441-4.128 2.765-13.678 9.614-29.577 2.015l1.869 1.797c0.699 0.63 1.554 1.362 2.481 2.077 11.418 8.53 23.62 7.303 32.769 1.243 1.267-0.838 2.424-1.609 3.507-2.334v-12.234m0-24.61c-10.02 6.738-23.546 15.833-26.085 17.536-4.127 2.765-13.82 9.708-29.379 2.273l1.804 1.729c0.205 0.19 0.409 0.375 0.612 0.571l-0.01 0.01 0.01-0.01c12.079 10.22 25.379 8.657 34.501 2.563 5.146-3.436 12.461-8.38 18.548-12.507l-0.01-12.165m0-24.481c-14.452 9.682-38.162 25.568-41.031 27.493-4.162 2.789-13.974 9.836-29.335 2.5l1.864 1.796c1.111 1.004 2.605 2.259 4.192 3.295 10.632 6.792 21.759 5.591 30.817-0.455 6.512-4.351 22.528-14.998 33.493-22.285v-12.344";
```
lindsaycarr: you should also declare `var wave =` here, too. Unless `wave` exists somewhere else, too.

ldecicco-USGS:
for the national map:
image

For 1 state:
image

Lindsay tells me the fill isn't black for her...I'm not sure what's up with my computer.

Total time: 10 minutes to add the path....4 hours to get it to link to the usgs.gov page correctly.

js/watermark.js
```diff
@@ -0,0 +1,54 @@
...
+  
+  var watermarkHeight = 100,
+      watermarkWidth = 200;
+  var bufferFromEdge = 25;
+  
+  // increase minimum so that you can actually see it
+  function precisionRound(number, precision) {
+    var factor = Math.pow(10, precision);
+    return Math.round(number * factor) / factor;
+  }
```
lindsaycarr: I think you're safe to delete lines 7-11. That function was used in legend to round the number for the size of the circles.

js/watermark.js
```diff
@@ -0,0 +1,54 @@
...
+        .attr("xlink:href", "https://www2.usgs.gov/water")
+        .attr("target", "_blank");
+
+    watermark
+      .append('rect')
+        .attr("x", 0)
+        .attr("y", 0)
+        .attr("width", 500)
+        .attr("height", 140)
+        .attr("fill", "white");
```
jread-usgs: do we want the box to have a white fill? What if only the wave and the USGS text had opacity?

js/watermark.js
```diff
@@ -0,0 +1,54 @@
...
+        .append('a')
+        .attr("xlink:href", "https://www2.usgs.gov/water")
+        .attr("target", "_blank");
+
+    watermark
+      .append('rect')
+        .attr("x", 0)
+        .attr("y", 0)
+        .attr("width", 500)
+        .attr("height", 140)
```
lindsaycarr: You have `watermarkHeight` and `watermarkWidth` as variables at the top, so I think you'd probably want to use them here? Also, x and y should probably also use the `xPosWatermark` and `yPosWatermark` so that you only need to change the numbers in one place if you want to move it

js/build_map.js
```diff
@@ -100,6 +100,9 @@ function create_map() {
   // add legend
   addLegend(minWateruse, maxWateruse);
   
+  // add watermark
+  addUSGS();
```
lindsaycarr: can we just change to `addUSGSWatermark` or `addWatermark`? Might add something else USGS-y and then this will be confusing?

js/watermark.js
ldecicco-USGS: I was thinking for the state view it's needed (?)

lindsaycarr: well, we want to be able to see it when you zoom in and it's over top of some circles? Would just having opacity do that?

js/build_map.js
ldecicco-USGS: sure

js/watermark.js
ldecicco-USGS: whoops

js/watermark.js
jread-usgs: In my opinion, no, we'd want only the wave and text to be seen, and it is OK for it to be subtle.

ldecicco-USGS: image

?

lindsaycarr: ☝️ if we are doing this, I think you can get rid of the rectangle entirely and just adjust the size using `scale(0.25)` as you have done already or add a height and width to the group.

ldecicco-USGS: The rectangle makes it so the user doesn't have to hit the letters exactly (so, the link is only for what's within the path). I'll defer to @jread-usgs for whatever behavior he wants

jread-usgs: I'd say drop the rect and that way the users can still mouse/tooltip things than may be partially overlapped.

jread-usgs:
Should we take the `watermark` fetcher out of vizlab? or do we need to update it? https://github.com/USGS-VIZLAB/vizlab/blob/a30cea3fdf0264dec087f4f1ff5182d3ebe6ba15/R/fetch.R#L89

ldecicco-USGS:
I think we could merge this and put in the backlog issues to use the "fetch.vizlab" svg if we want...

Issue #15, opened 2018-02-26
Add analytics for which states/categories/stories are viewed


aappling-usgs:

lindsaycarr:
Need to wait for interface (buttons, etc) to be a little more finalized before starting this task.

wdwatkins:
Is this ready to go yet?

aappling-usgs:
I think so! Would be great if you could take it on.

ldecicco-USGS:
Reminder to me (and maybe others):
https://developers.google.com/analytics/devguides/collection/gtagjs/migration
That's what we've moved to.

ldecicco-USGS:
Wondering if we'll want to (and it's reasonable to?) track the hovers over the categories. Personally, I only hover, so only recording clicks on the categories might not tell us much.

Other than that, we want to see when people went from national -> (which) state, and (which) state -> (which) state, and (which) state -> national

Once we see the stories we can think about the other stuff.

wdwatkins:
we tracked state hovers for the first water use viz

wdwatkins:
Hmm, I can't speak to performance impacts, but I wouldn't think it matters?  Googling it seems inconclusive, but everything should be loaded by that point, so I wouldn't think there is anything to hold up?

jread-usgs:
Agreed that this was useful info last time, but also note that David made some special code so that it didn't end up triggering too many events for the hovers.

aappling-usgs:
@wdwatkins what is left to do on this issue?

wdwatkins:
We could add scroll depth tracking for key stories, I've not implemented that before though.  There are a few tweaks that could be made to what I submitted in my PR also (mainly the namespacing, but we could also record the previous category in the category change event.

aappling-usgs:
* Scroll depth tracking for key stories sounds like a great idea. I think we've done that in previous vizzies - hopefully there's someone around who remembers.
* Recording the previous category change seems like a pretty easy change, right? If so, I'm in favor.
* I added issue #127 to deal with namespacing for all sorts of global variables all at once. Let's keep that one as a separate issue from this one.

aappling-usgs:
I've opened #146 to cover scroll depth tracking. Recording the previous category is getting done in #137 (https://github.com/USGS-VIZLAB/water-use-15/pull/137/files#diff-2afd72a00f31d3d53312dce9ff03a875R177). So I think we're good for basic analytics and can follow up with #127 and #147 to refine.

[...]

PR #125, opened 2018-04-03
figure analytics

wdwatkins:
starting on #15 

I'm recording category and view for all events.

I'm not 100% sure on the functioning of the category changes and their corresponding events, so some extra attention in review would be warranted there.  I think there may be some odd behaviour there as to how the new category data is loaded as well, see my comments in #97

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -32,7 +32,7 @@ function addStates(map, stateData) {
   var newView = getHash('view');
   if(newView == null) { newView = 'USA'; }
   if(newView != 'USA') {
-    updateView(newView);
+    updateView(newView, fireAnalytics = false);
   }
 }
 
@@ -106,7 +106,7 @@ function zoomToFromState(data) {
   updateView(newView);
 }
 
-function updateView(newView) {
+function updateView(newView, fireAnalytics = true) {
   // update the global variable that stores the current view
   activeView = newView;
   
@@ -175,7 +175,12 @@ function updateView(newView) {
       "translate(" + chart_width / 2 + "," + chart_height / 2 + ")"+
       "scale(" + zoom_scale + ")" +
       "translate(" + -x + "," + -y + ")");
-  
+      
+  if(fireAnalytics) {
+    gtag('event', 'update view', {
+  'event_category': 'figure',
+  'event_label': newView});
+  }    
 }
 
 function updateCategory(category, prevCategory) {
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -108,6 +108,7 @@ function zoomToFromState(data) {
 
 function updateView(newView, fireAnalytics = true) {
   // update the global variable that stores the current view
+  oldView = activeView;
   activeView = newView;
   
   // update page info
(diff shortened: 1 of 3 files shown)


js/map_utils.js
```diff
@@ -175,10 +176,18 @@ function updateView(newView) {
       "translate(" + chart_width / 2 + "," + chart_height / 2 + ")"+
       "scale(" + zoom_scale + ")" +
       "translate(" + -x + "," + -y + ")");
-  
+  // don't need timeout for view change   
+  if(fireAnalytics) {
+    gtag('event', 'update view', {
+  'event_category': 'figure',
+  'event_label': 'newView=' + newView + '; oldView=' +     oldView + '; category=' + activeCategory
```
wdwatkins: `activeCategory` here isn't passed into this function.  Is it bad practice in js to use variables in higher-level environments?

aappling-usgs: yes and no. i think we're supposed to be using a namespace, and we're not yet, but i haven't figured out how to do that yet. on the other hand,in a responsive environment like a web page, i think it's reasonable to have variables that don't get passed as arguments. this would be a good one to run past mary, maybe next week or late this week

aappling-usgs: and :+1: for pointing this out

aappling-usgs:
Hmm - i approved and then re-noticed your appeal for extra review on the category changes. I can think harder about that tomorrow, or maybe @lindsaycarr would be able to make a quicker judgment than I on that part. Either way is fine with me - if @lindsaycarr approves, I do too, or else I can look more once I get in the office 2 hours behind you.

js/map_utils.js
```diff
@@ -187,7 +196,15 @@ function updateCategory(category, prevCategory) {
...
   
   updateLegend(category);
-  
+  if(updateCategoryTimer){
+    clearTimeout(updateCategoryTimer);
+  }
+  updateCategoryTimer = setTimeout(function(){
+     gtag('event', action + ' update category', {
+  'event_category': 'figure',
+  'event_label': category + '; view=' + activeView
```
lindsaycarr: would you want to use `prevCategory` in the event tag? Noticed you did `oldView` and `newView` in the other one.

js/map_utils.js
```diff
@@ -226,12 +245,23 @@ function showToolTip(d, category) {
...
+  }, toolTipDelay);
 }
 
 function hideToolTip() {
   d3.select(".tooltip")
     .classed("shown", false)
     .classed("hidden", true);
+  if (toolTipTimer){
+      clearTimeout(toolTipTimer); // stop ga for edge states
+    }
```
lindsaycarr: not sure what you mean by "stop ga for edge states". Wouldn't it just get cleared the next time `showToolTip` is called?

js/map_utils.js
wdwatkins: It seemed less valuable to me than knowing the previous view.  Although I suppose there's no reason not to include it other than my own laziness 😁

js/map_utils.js
```diff
@@ -187,7 +196,15 @@ function updateCategory(category, prevCategory) {
   updatePies(category, prevCategory);
   
   updateLegend(category);
-  
+  if(updateCategoryTimer){
+    clearTimeout(updateCategoryTimer);
+  }
+  updateCategoryTimer = setTimeout(function(){
```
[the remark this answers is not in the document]
wdwatkins: Possibly, it depends how much that delay turns out to be.  We just want try to collect events when the user is actually looking at a category, not if they just incidentally drag the cursor over something.

js/map_utils.js
lindsaycarr: ahhhh OK

js/map_utils.js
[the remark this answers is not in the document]
lindsaycarr: right and I think that's exactly what we want for mouseovers - delay so that it only triggers when someone is actually hovering over that category wanting to inspect it and not accidentally passing over, but I could see still adding some extra delay to the ga event on top of the delay for the category switch

Issue #95, opened 2018-03-26
Show adjacent states when zoomed in (S)


aappling-usgs:
Just show state boundaries/background, not counties, for non-focal states.

Design decision: How should focal vs non-focal states be indicated - by position, by varying shades of grey in background, by the presence/absence of county boundaries, etc.?

jread-usgs:
For starters, the adjacent states wouldn’t display counties and fill would be a different (lesser emphasis) shade. Might want to consider also using a different border style even if simple as thicker stroke compared to neighbors.

lindsaycarr:
PR coming soon for this!

Moved counties under states boundaries to get rid of this:
![image](https://user-images.githubusercontent.com/13220910/38276705-847600c4-375b-11e8-963d-18d5fa011750.png)

Tried to get county lines to transition to turn on, but couldn't get them to work unless I commented out the transition part (link to commit: ). Also, had to make highlighting a county on hover be in js if emphasizing a fill for the zoomed in state was also in js (otherwise emphasized fill color trumped my css class and you couldn't see what you hovered over). So, in the end, reverted back to all of those style changes being class switches (so no transitioning). The added benefit of this method is that it's actually much faster when you're hovering on and off of counties than if we switched their fill via js. 

Tried dotting adjacent state lines rather than just filling focal state. Worked with TX because it's one of the last paths, but didn't work with AL because adjacent states are painted on after:

![image](https://user-images.githubusercontent.com/13220910/38326939-a84fe06c-380c-11e8-9037-b8a07dcce6e9.png)

![image](https://user-images.githubusercontent.com/13220910/38326997-c53f4b04-380c-11e8-8557-88d9aabbc6b6.png)

aappling-usgs:
What if the focal state gets a thicker black border than the dashed lines, such that it doesn't matter which is on top of which?

lindsaycarr:
Yeah, that's what was happening. Although, I don't think I was scaling the non-focal state borders like the focal state was...which would explain why TX is fine. I'll try that

aappling-usgs:
I think this is done. Lindsay spent 5 hours.

PR #129, opened 2018-04-04
Adjacent states view

lindsaycarr:
#95, comments there.

neighborpanning

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -49,34 +49,6 @@ formatState = function(attr, d, active) {
   return stateStyle[view][activeness][attr];
 }
 
-// on mouseover
-function highlightState(selection) {
-  selection
-    .style('fill', function(d) { return formatState('fill', d, true); })
-    .style('stroke', function(d) { return formatState('stroke', d, true); })
-    .style('stroke-width', function(d) { return formatState('stroke-width', d, true)/zoom_scale; });
-}
-
-// on mouseout
-function unhighlightState(selection) {
-  selection
-    .style("fill", function(d) { return formatState('fill', d, false); })
-    .style('stroke', function(d) { return formatState('stroke', d, false); })
-    .style('stroke-width', function(d) { return formatState('stroke-width', d, false)/zoom_scale; });
-}
-
-// on mouseover
-function highlightCounty(selection) {
-  d3.select(selection)
-    .style('fill', function(d) { return "darkgrey"; });
-}
-
-// on mouseout
-function unhighlightCounty(selection) {
-  d3.select(selection)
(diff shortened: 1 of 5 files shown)


lindsaycarr:
@aappling-usgs tried to bring back dashed borders for background states, but some of the lines just look solid...

scaling_dashes

aappling-usgs:
Nah, I prefer thin solid lines over dashed lines anyway.

lindsaycarr:
So go back to all solid?

lindsaycarr:
Might wanna let our design lead take a look :)

aappling-usgs:
I do prefer solid to dashed and thought that was what you had going in this PR based on your GIF.

And yes, let's get JR in on this - _Design lead @jread-usgs, yoo hoo!_

js/map_utils.js
```diff
@@ -146,27 +115,33 @@ function updateView(newView, fireAnalytics = true) {
...
-    states
-      .transition()
-      .duration(750)
-      .style("fill", function(d) { return formatState('fill', d, false); })
-      .style("stroke", function(d) { return formatState('stroke', d, false); })
-      .style("stroke-width", function(d) { return formatState('stroke-width', d, false); });
-  } else {
-    showCountyLines(activeView);
-    states
+  // set the styling: setting by adding or removing class, so d3 transitions not used
```
aappling-usgs: is this an aesthetic or performance decision? i agree that the aesthetics of doing it with classes are fine, but i don't necessarily prefer them to the d3 option, so just wondering why you switched.

lindsaycarr:
That's what I had initially pushed, then your comment in #95 made me think to try out dashed again so that's what the last two commits of this have.

jread-usgs:
silently lurking. I prefer solid lines too.

lindsaycarr:
Alrighty - reverting! Anything else major that should be changed with this one?

aappling-usgs:
Nothing from me. Looking forward to resolving the merge conflicts between yours and mine in a minute =)

lindsaycarr:
Made the lines solid and a little thicker/thinner. 

image

js/map_utils.js
lindsaycarr: I explained it in more detail here https://github.com/USGS-VIZLAB/water-use-15/issues/95#issuecomment-378701304, but some was performance and some was actually getting the styles to work correctly by having them here.

lindsaycarr:
Oh and this was ~5 hrs. Not much code, but a lot behind-the-scenes of thinking where everything should live. Also, struggled with why `.transition().duration(750).style("fill"...` wasn't doing anything for awhile before finally going back to just changing the css class.

PR #130, opened 2018-04-04
Split spatial data

### LINKED ISSUE #111: Split county & state boundary datasets into coarse and fine resolutions
We should be able to have one national-scale, coarse-resolution dataset for the national view, then one (or 52?) state-scale, finer-resolution dataset[s] for when you zoom in.

aappling-usgs:
Resolves #111, Split county & state boundary datasets into coarse and fine resolutions

We start by loading and displaying a low-res set of state boundaries. Then we jump right to loading county boundaries to support interaction. Then when you zoom in, we load and display high-res county and state boundaries.

This should speed up the initial page load time because we're loading smaller files to begin with.

We could play more with the specifics, but for now I think I have something adequate, which is:
* when zooming in, only the state boundaries (nationally) and the county boundaries specific to the focal state are updated. This leaves some funniness where state boundaries and county boundaries of adjacent states are mismatched, but that issue should go away once we incorporate @lindsaycarr 's open PR (https://github.com/USGS-VIZLAB/water-use-15/pull/129).
* Then if you pan to an adjacent state, that state's county borders are updated on the fly. I think all that is sensible and efficient.
* When you zoom back out, we revert to coarse-resolution borders. This may not be the best option. The data is loaded and cached the first time you zoom in, so it's not that we're wasting time loading the data again next time you zoom it, but it does take some processing power each time we switch the county and state boundary DOM elements between low and high resolution. We could experiment with keeping the high-resolution data applied forever once it's loaded; we'd just want to look at how that affects interaction speed at the national scale.

We could also consider ONLY doing lo-res on mobile, even when zoomed in. The counties are sorta blocky with the current settings (`simplify=I(1e-6), quantize=I(1e6)` in preprocess.yaml, line 51), but mobile screens are small.


aappling-usgs:
OK, I think that merge & resolution worked. Testing it locally, no problems.

When you finish reviewing this, could you give me some time before approval and merge to make one more commit? I'd like to do the reverse of this: https://github.com/USGS-VIZLAB/water-use-15/pull/130/commits/19e906a1b70f79acd4186ba069c2c62384badcae so that our working datasets are not prefixed by `test_` once this gets merged.

scripts/preprocess/process_compute_centroids.R
```diff
@@ -27,5 +27,5 @@ process_compute_centroids <- function(outfile, spdat, county_dict_file) {
   centroid_spdf <- sp::SpatialPointsDataFrame(centroids, spdat@data, match.ID=FALSE, proj4string=proj4string(spdat))
   
   # write to file
-  geojsonio::topojson_write(centroid_spdf, file=outfile, overwrite=TRUE) # geojson would be about 37% larger
+  geojsonio::topojson_write(centroid_spdf, file=outfile, overwrite=TRUE, object_name='centroids') # geojson would be about 37% larger
```
lindsaycarr: 👍 OH - didn't know that trick!

scripts/process/merge_centroids_wu.R
```diff
@@ -1,17 +1,18 @@
...
     rowwise() %>% 
     mutate(other = total - sum(thermoelectric, publicsupply, irrigation, industrial))
   
+  # put the result back into the topojson
   topo@data <- county_data
   
-  # write to file
-  geojsonio::topojson_write(topo, file=viz[['location']])
+  # write to file. don't use geojsonio::topojson_write because we can't specify two layers [yet; geojsonio is growing fast]
+  geojsonio::topojson_write(topo, file=viz[['location']], overwrite=TRUE, object_name=viz$process_args$layer)
```
lindsaycarr: so your comment on the line above this one says "don't use" but we are still using `geojsonio::topojson_write`. Is there an alternative we should be using here?

js/map_utils.js
```diff
@@ -19,27 +21,28 @@ function addStates(map, stateData) {
...
   if(newView != 'USA') {
     updateView(newView, fireAnalytics = false);
   }
 }
 
 // Function to look up a style
-formatState = function(attr, d, active) {
+function formatState(attr, d, active) {
+  var view;
   if(activeView == 'USA') {
```
lindsaycarr: does this need `===` instead of just two?

js/map_utils.js
```diff
@@ -19,27 +21,28 @@ function addStates(map, stateData) {
...
-  if(newView == null) { newView = 'USA'; }
+  if(newView === null) { newView = 'USA'; }
   if(newView != 'USA') {
     updateView(newView, fireAnalytics = false);
   }
 }
 
 // Function to look up a style
-formatState = function(attr, d, active) {
+function formatState(attr, d, active) {
```
lindsaycarr: I am trying to see if your PR uses this function, but I ended up completely getting rid of this. I'm not seeing a place in your PR that would still need this either.

js/map_utils.js
```diff
@@ -115,6 +118,10 @@ function updateView(newView, fireAnalytics = true) {
       nationDims.width/stateDims.width]);
   }
 
+  // update the geospatial data for the upcoming resolution
+  updateCounties(activeView);
+  updateStates(activeView);
+
```
lindsaycarr: I can't add a comment there, but on a line below within `updateView`, your merge conflicts I think kept these lines and shouldn't have. That's wrapped up in `backgroundState` and `foregroundState` functions but having it here is overriding those.

```
statecounties
      .transition()
      .duration(500) 
      .style("stroke-width",  0.75/zoom_scale); // make all counties have scaled stroke-width
```

scripts/preprocess/process_compute_centroids.R
aappling-usgs: it's new as of about last week! Scott Chamberlain has been working on the package. it's on CRAN, though

scripts/process/merge_centroids_wu.R
aappling-usgs: oh, right. i tried several options, but i think this one is the best. i was trying to get states and counties into the same file, with water use data, but ultimately had to give up. i tried reading and writing using jsonlite, but that was corrupting my topojson in ways I couldn't find or fix. and as noted in that comment (which i'll delete), with the option i chose i've had to go back to having separate files for states and counties - a shame, but also should allow us to load the states before the counties for faster time to viewable content.

js/map_utils.js
aappling-usgs: yes, thanks

js/map_utils.js
aappling-usgs: oh, i guess that's a merge issue. taking out.

js/map_utils.js
aappling-usgs: so just delete the lines you quoted, right?

aappling-usgs: hmm - are the county lines supposed to be giant now?
image

aappling-usgs: now we have

image

which looks more like what i expected/wanted

lindsaycarr: But what I'm not seeing is the state border getting much thicker than the background state borders....

lindsaycarr: I expect this, where foreground state is bold and background states aren't. I should've looked more closely at the code I suggested to throw out, but there is still something different

image

js/build_map.js
```diff
@@ -103,11 +103,12 @@ function create_map() {
   // add watermark
   addWatermark();
   
-  // add placeholder group for county boundaries
-  map.append('g').classed("county-bounds", true);
+  // add placeholder groups for state and county boundaries
+  map.append('g').attr('id', 'state-bounds');
```
lindsaycarr: I think this is the difference for why we can't see foreground states as bold. I added counties underneath the state bounds

lindsaycarr: Just tested on your branch that I pulled down. If you just switch these lines so that the county group is added first, then it should work!

aappling-usgs:
OK @lindsaycarr , I think we're there! I also just ran the preprocessing step again to push the revised/renamed files to SB so others can use them

aappling-usgs:
Oh...except the conflicts, which I thought I just resolved...

aappling-usgs:
*now* it's ready

PR #134, opened 2018-04-05
remove legend

mwernimont:
Removed the function call from the build_map.js and scrapped the legend_utils.js. No mo legend


ldecicco-USGS:
https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L101

mwernimont:
@aappling-usgs good call, found it and deleted it.

js/map_utils.js
```diff
@@ -170,10 +170,6 @@ function updateCategory(category, prevCategory, action) {
   
   updatePies(category, prevCategory);
   
-  updateLegend(category);
-  if(updateCategoryTimer){
-    clearTimeout(updateCategoryTimer);
-  }
```
aappling-usgs: I think these three lines 
```
if(updateCategoryTimer){
  clearTimeout(updateCategoryTimer);
}
```
are relevant to google analytics rather than to the legend, aren't they? and so shouldn't be deleted? or am i misunderstanding their function? @wdwatkins ?

mwernimont: oh I see now, my bad.

mwernimont:
Apologies just saw the stacked code and thought it was all the function.

Issue #126, opened 2018-04-04
Streamline pointer events


aappling-usgs:
From David in https://github.com/USGS-VIZLAB/water-use-15/issues/97#issuecomment-378413238:

Looks like we reload a category whenever the button is hovered over, even if that category is already loaded. Same with the mouseout. There should be a check there to see if the category is already loaded.

And there seems to be mouse-in/mouse-out happening when you click on a category, not sure that should be happening. I would think it is less snappy since it is loading the new circles twice.

Might we also want to add a very small delay to deciding to load a new category, so just incidentally dragging the cursor over the buttons doesn't trigger category changes? Although if the category changes get snappier this might not be an issue.

Lindsay says:

that's a good idea....Here's an example for click vs double click and I think that's pretty much what we would use for mouseover vs click. We should also handle mouseout differently if it's been clicked. No need to revert the category.

https://github.com/almende/vis/issues/203#issuecomment-48481501

wdwatkins:
To further clarify, I would suggest handling the mouse in/out behavior before adding any delay with the category change.  IMO the delay is only needed if there is still a substantial performance penalty for switching categories.

lindsaycarr:
Well, I think we need some type of delay regardless so that moving your cursor just over the top accidentally doesn't initiate a switch. It should definitely be lower priority than the other changes, and should be a pretty small delay, like 100 ms or something.

So things to fix:

1. Don't switch categories if moused or clicked on the same one (handle [here](https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L140-L146)).
2. Don't revert categories if mousing out of the same one that is already active (handle [here](https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L148-L150)).
3. Add small delay to mouseover events so that accidental hovers don't trigger a category switch (handle [here](https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L145-L147)).

aappling-usgs:
Tasks 1 and 2 are done. I'm OK punting on task 3, which may or may not be important as we improve the load time.

PR #137, opened 2018-04-05
starting on streamlining pointer events

wdwatkins:
There are three items to fix from #126:

1. Don't switch categories if moused or clicked on the one that is already active (handle here).
2. Don't revert categories if mousing out of the one that is already active (handle here).
3 .Add small delay to mouseover events so that accidental hovers don't trigger a category switch (handle here).

This handles 2.  But 1 and 3 aren't quite there yet.  

Once a category is clicked, there are no issues with it.  But if you hover on a category, then click, it is still loaded twice.  I think the delay is working on the mousein, but it needs to also catch the mouseout event, which isn't happening yet.


js/build_map.js
```diff
@@ -140,12 +142,18 @@ var categoryButtons = d3.select('#button-container')
   .on('click', function(d){
     prevCategory = activeCategory;
     activeCategory = d.toLowerCase(); // put this here so it only changes on click
-    updateCategory(activeCategory, action = 'click');
+    updateCategory(activeCategory, prevCategory.toLowerCase(), 
+    action = 'click');
```
aappling-usgs: I'm wondering if we should be dividing the ideas in `updateCategory` into actions that need to happen when we preview a category and those that only happen when we click and commit to a category. For example, I believe we don't even need to be updating the category display (the circles) during calls to `on('click')` - by the time a mouse gets to clicking on a button, hasn't it already been hovering, and therefore aren't we already always viewing the category being clicked? So maybe `updateCategory` only needs to update the title, hash, category info, etc., like this: 

```
function updateCategory(category, prevCategory) {
      // update the globals about category view status
      prevCategory = activeCategory;
      activeCategory = d.toLowerCase();

  // update page info
  updateTitle(category);
  setHash('category', category);

  documentCategorySwitch(category, prevCategory, action='click');
}
  
function showCategory(category, prevCategory, action) {
  documentCategorySwitch(category, prevCategory, action);
} 

function documentCategorySwitch (category, prevCategory, action) {
  if(updateCategoryTimer){
    clearTimeout(updateCategoryTimer);
  }
  updateCategoryTimer = setTimeout(function(){
     gtag('event', action + ' update category', {
  'event_category': 'figure',
  'event_label': category + '; view=' + activeView
     });
  }, updateCategoryDelay);
}
```

where `.on('click')` calls updateCategory whereas `.on('mouseover')` and `.on('mouseout')` call `showCategory`

I'm just drafting, not testing, this code, but it might be a start on finishing out task 1 - what do you think?

wdwatkins: Yeah, that sounds good

aappling-usgs: yes, good point. maybe we make it two different gtag calls, one in updateCategory (without a timer) and one in showCategory (with a timer).

js/build_map.js
```diff
@@ -137,12 +139,11 @@ var categoryButtons = d3.select('#button-container')
   .on('click', function(d){
     prevCategory = activeCategory;
     activeCategory = d.toLowerCase(); // put this here so it only changes on click
```
aappling-usgs: looks like the above two lines may now be redundant with map_utils.js lines 159-160? i think you could take out these two here, then pass `d.toLowerCase()` to updateCategory (or do the lowercasing within updateCategory, whichever)

js/build_map.js
```diff
@@ -137,12 +139,11 @@ var categoryButtons = d3.select('#button-container')
...
-    updateCategory(activeCategory, action = 'click');
+    updateCategory(activeCategory, prevCategory.toLowerCase());
```
aappling-usgs: do we need to call toLowerCase on prevCategory? i think we're doing a pretty good job of ensuring that prevCategory is always lowercase to begin with, since it either comes from activeCategory or is set to 'total'

aappling-usgs:
Please add an hours estimate to https://github.com/USGS-VIZLAB/water-use-15/issues/15#issuecomment-379386161 when you're ready

PR #138, opened 2018-04-05
remove viz.yaml legend tidbits

lindsaycarr:
follow up to #134, just missed the pieces in the `viz.yaml` which caused the build to fail.


aappling-usgs:
oh, yep, i just did that locally too.

PR #144, opened 2018-04-06
svg within a key story div

### LINKED ISSUE #142: How to source javascript that can be used w/in key stories
Question - 

If we want to use d3 in a key-story div, how does that happen? The way the page is set up, our div isn't created when the functions would be called. I chatted w/ Lindsay about this yesterday, and she pointed out [this reference](https://stackoverflow.com/questions/16823757/d3-selectelement-not-working-when-code-above-the-html-element) which covers what is going on. 

For example, if I add this snippet to the build_map.js (note I would do this stuff in a different file, but bear with me...):
```js
var svgStates = d3.select(".key-story-plot")
  .append("svg")
  .attr('viewBox', '0 0 1000 300')
  .attr('preserveAspectRatio', 'xMidYMid');
svgStates.append("g")
  .append('rect')
    .attr('fill','red')
    .attr('height',"300")
    .attr('width',"1000");
```
 I would expect a red rectangle to appear in the first key story div. It doesn't, but it will appear if I run that same snippet in console after the page is loaded. 

I know they key story divs don't appear the same way after the recent PR was merged, but this question is still relevant. I can work around this, but wondering what our preferred method for sourcing function calls that depend on DOM elements like these divs. The build_map gets around this by creating the first div of the page by using d3.

jread-usgs:
This is a small PR to capture the pattern adding a few function calls for content w/in a key story div.

fixes #142 if this is acceptable.


jread-usgs:
@aappling-usgs I thought I followed the suggestion. What did I miss?

viz.yaml
```diff
@@ -302,6 +308,7 @@ publish:
   -
     id: key_stories
     template: layout/templates/key_stories.mustache
+    depends: "rank_states_utils"
```
aappling-usgs: below this line add
```
    context:
      sections: ["rank_states_utils"]
```

viz.yaml
```diff
@@ -242,7 +243,8 @@ publish:
       resources: ["lib_d3_js", "topojson", "header_css", "content_css", "footer_css", "socialCSS",
                   "map_css", "custom_css", "hashes", "projection", "styles", "map_utils", "map_states", "map_counties", 
                   "watermark_utils", "circle_utils"]
-      sections: ["social","map", "directions_text", "key_stories"] # adding header and footer here, doubled each on the page
+      sections: ["social","map", "directions_text", "key_stories", "rank_states_utils"] # adding header and footer here, doubled each on the page
```
aappling-usgs: remove rank_states_utils from here (put it below instead)

viz.yaml
```diff
@@ -227,6 +227,7 @@ publish:
       projection: "custom_projection"
       styles: "styles"
       map_utils: "map_utils"
+      rank_states_utils: "rank_states_utils"
```
aappling-usgs: remove rank_states_utils from here (can just be listed as a depends of the key_stories viz item)

layout/templates/key_stories.mustache
```diff
@@ -57,7 +57,9 @@ Generating electricity uses large amounts of water. In most large <b>thermoelect
     
     <p>
 As a result of all this variation in how we use water, total water use differs greatly among statesis not simply proportional to population. Can you guess how your state ranks in total water use and use by category?
-</p>
+    </p>
+    <div class="key-story-plot" id="rank-states-interactive"></div>
+
```
aappling-usgs: somewhere in this file add 
```
{{#sections }}
    {{{ . }}}
{{/sections }}
```

lindsaycarr:
@aappling-usgs yes, this would do more of what I was talking about. Although, I think that the `.js` file needs to be listed somewhere as a `resource` or it won't get copied into target? Might be wrong about that though

aappling-usgs:
Cool. I think it's getting copied because it's listed as a `depends` of `key_stories` (line 311). I think including something in `resources` causes that thing to be loaded in the head of the html file (in a script tag, e.g., <script src="js/states.js?_c=6749086" type="text/javascript"></script>). This merged PR seems to be working both for Jordan and for me locally. The publish stuff is still a bit black-boxy to me, so maybe I'm getting the interpretation wrong, but I think that's what's happening.

PR #145, opened 2018-04-06
update after review

jread-usgs:
for completeness


aappling-usgs:
Cool, and it works like you wanted?

jread-usgs:
all good here, yes.

PR #148, opened 2018-04-07
Circles as paths

aappling-usgs:
Don't merge yet - I'm stuck on something and looking for help.

Successes:
* Circles are defined as a single path per 
* Highlights for one county at a time still work (different implementation)
* Centroid and county datasets are both smaller for faster load

Trouble:
The Total circles start by showing as the correct blue, but then they snap into only showing circle strokes, not fill. I have no idea why - the properties sure look like fill=stroke=blue.

180407-circle-paths

Any ideas?


jread-usgs:
That’s really odd. Have you tried using a different color for total just to see if it is barfing on the end value rgba as opposed to something specific to total paths?

aappling-usgs:
Yes, I have tried using the same color as `irrigation` or `other` and it does the same thing. I'm moving toward the hypothesis that it happens when there are excess demands on the animation system - somehow the transitions just break down and stop partway through rather than jumping to the correct end state. But I don't know how to further test or address that possibility.

Other things I've learned:
* If I zoom out while in Total view, somewhere around 75% zoom I get filled blue circles back.
* If I zoom back in from scale where filled circles are visible, somewhere between 80% and 125% (it varies) I get empty blue circles back.
* I can sometimes generate empty circles in other categories - for example, in Irrigation view when zoomed in and switching between states. See GIF below.
* The failures seem to sometimes, but not always, be associated with `[Violation] 'click' handler took 256ms` in the console (violations between ~250 and ~300 ms) or `[Violation] 'requestAnimationFrame' handler took 50ms`

180408-circle-paths-fill-unfill

jread-usgs:
as a gut check, is the performance of the transitions an improvement over the branch that has them as circles?

aappling-usgs:
Yes, the transitions are lovely and smooth.

ldecicco-USGS:
FYI...in IE on dev, I don't even see a map due to the "zoomToFromState" being undefined...maybe that's still plaguing your local build on IE (and has nothing to do with the pr?)

aappling-usgs:
Thanks Laura. Yeah, I don't think the IE thing has to do with this PR, I was really just confirming that it's not just your computer =). But I'm also not seeing an error with zoomToFromState explicitly mentioned in the console, so I don't know why our IEs are failing in slightly different ways. Like you, I don't see a map.

Issue #149, opened 2018-04-09
Recover the favicon


aappling-usgs:
Could be a vizlab issue, could be a water-use-15 issue - needs some digging.

lindsaycarr:
@ldecicco-USGS can we get an hours estimate for this?

ldecicco-USGS:
30 minutes

PR #164, opened 2018-04-09
Adding favicon as resource

ldecicco-USGS:


ldecicco-USGS:
Fixes #149

Not sure what's going on exactly. The way we had it we could see a "wave" favicon building locally. I tried changing the Jenkins build to treat the .ico as an svg...but that didn't seem to solve the problem. This will add it specifically as a resource though, and that brings us back to a "vizlab" favicon.

ldecicco-USGS:
I don't think we want it in embed, I'd assume the developer of whatever site is using it will have their own.

ldecicco-USGS:
I'm making an issue still in vizlab because I thought we had it (or wanted it) that it would be in the main page template already as a resource. fullPage.mustache I think will do it

Issue #151, opened 2018-04-09
Troubleshoot IE


aappling-usgs:
It's just a big white space between the title and the button sright now

PR #166, opened 2018-04-09
Remove title

### LINKED ISSUE #158: Remove the title

ldecicco-USGS:
Fixes #158


aappling-usgs:
there must be a call or two to `updateTitle` somewhere, right? i think those will also need to be deleted.

ldecicco-USGS:
Bonus:
Fixes #151

Figured it out via:
https://stackoverflow.com/questions/15178700/javascript-functions-and-default-parameters-not-working-in-ie-and-chrome

js/map_utils.js
```diff
@@ -185,10 +186,22 @@ function documentCategorySwitch(category, prevCategory, action) {
 }
 
 
+function highlightCircle(currentCircle) {
```
ldecicco-USGS: I don't know what this is...

lindsaycarr: it's what makes the circle go from slightly see-through to solid when you hover

lindsaycarr: not sure how it is being added here though...they should've been there the whole time. Same with `unhighlightCircle` below

ldecicco-USGS: but how did it get in my pull request? probably something I did on the merge conflict

js/styles.js
```diff
@@ -29,14 +29,22 @@ function emphasizeCounty(selection) {
     .classed("emphasize-county", true);
 }
 
-function foregroundState(selection, scale = 1) {
+function foregroundState(selection, scale) {
```
lindsaycarr: so were these the trick to #151? Setting a default to the functions?

ldecicco-USGS: Seems like it

js/build_map.js
```diff
@@ -91,14 +95,12 @@ function prepareMap() {
 
   // create map title
   svg.append("text")
-    .attr("id", "maptitle")
     .attr("x", chart_width/2)
     .attr("y", chart_height*0.10); // bring in 10% of chart height
```
aappling-usgs: looks like the merge conflict resolution brought 3 of these 4 lines back in

js/map_utils.js
```diff
@@ -185,10 +186,22 @@ function documentCategorySwitch(category, prevCategory, action) {
...
+      origNode = orig.node();
+  var duplicate = d3.select(origNode.parentNode.appendChild(origNode.cloneNode(true), 
+                                                            origNode.nextSibling));
+                                                            
+  // style duplicated circles sitting on top
+  duplicate
+    .classed('circle-duplicate', true)
+    .style("pointer-events", "none")
+    .style("opacity", 1); // makes the duplicate circle on the top
+}
```
aappling-usgs: i don't think lines 190 through 200 should be in here following the merge of my PR

js/map_utils.js
```diff
@@ -185,10 +186,22 @@ function documentCategorySwitch(category, prevCategory, action) {
...
 
-function updateTitle(category) {
-  d3.select("#maptitle")
-    .text("Water Use Data for " + activeView + ", 2015, " + category);
+function unhighlightCircle() {
+  d3.select('.circle-duplicate')
+    .remove(); // delete duplicate
```
aappling-usgs: The unhighlightCircle function should look pretty different from both of these. Should be: https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/circles.js#L100-L105
```js
function unhighlightCircle() {

  map.select('circle#wu-highlight')
    .classed('hidden', true);

}
```

js/map_utils.js
aappling-usgs: oh! also, the functions highlightCircle and unhighlightCircle should now live in circles.js, not here in map_utils.js

js/map_utils.js
aappling-usgs: yes, i think this is a merge thing, probably because i moved highlightCircle and unhighlightCircle to circles.js

aappling-usgs:
i think you still need to delete lines 97-99 of build_map.js

js/build_map.js
```diff
@@ -91,14 +91,12 @@ function prepareMap() {
...
   
   /** Initialize title and URL **/
```
ldecicco-USGS: So...these comments need to be deleted?

PR #169, opened 2018-04-10
footer thumb alt text

wdwatkins:
This requires https://github.com/USGS-VIZLAB/vizlab/pull/365 to show up, since the alt text was not being rendered in the footer template.  However, it shouldn't break the build to not have the vizlab update.


wdwatkins:
I _think_ these are the only missing alt texts, and that the other errors are related to some problems with `<a> tags.

Issue #140, opened 2018-04-05
Implement mobile drop down selector (M)


jread-usgs:
- In the state view, our dropdown shifts a little to make room for zoom out button 
 - in the state view, a second (smaller text) dropdown appears for choosing counties. At the state-only (no selection) view, no county is selected
 - selecting a county from the second dropdown causes the county to be highlighted, changes the selection name in #139 but does not alter any other elements of the view
 - zoom out takes you back to the US view and removes any selection of a county. 
 - options in the secondary drop down only includes counties within the active state. 

National view:
![image](https://user-images.githubusercontent.com/2349007/38381899-e9cfcb26-38cd-11e8-8ffc-6ca19c20f2bd.png)

State view:
![image](https://user-images.githubusercontent.com/2349007/38381867-c6a22cfc-38cd-11e8-9468-dcf196824cf5.png)

Issue #139, opened 2018-04-05
Implement new category selector/legends (M)


jread-usgs:
This task is to change the buttons:
 - The selected button will expand to the right and all other buttons will change to a neutral grey
 - the selected button will contain "{view_selection}.XX" for the time being (XX filled in for another task). `view_selection` is `us`, `st`, `cty` as a placeholder to make sure we are aware the data will change with each selection. 
 - In desktop, the buttons will be on the left side of the map
 - Desktop will have a title for the categories which will adjust based on the selection (no title for mobile)
 - Mostly opaque backing rectangle for the desktop view
 - in mobile, the buttons will be under the map
 - Am assuming we want the buttons to be integrated within the svg, not outside of it (current implementation)
 - This does not include the drop down selector tasks
 - Note for desktop, this change likely impacts the math behind our state zoom, since we won't want the category/legend to sit on part of the active state. Dealing w/ this issue is outside of the scope of this task. 

See mobile here: 
![image](https://user-images.githubusercontent.com/2349007/38381709-401be4c0-38cd-11e8-8cba-ad561e584b58.png)
and desktop here:
https://github.com/USGS-VIZLAB/water-use-15/issues/89#issuecomment-378579693

jread-usgs:
I didn’t think we were planning to have the county drop down or the zoom out visible at the national scale. We’ve got some white space above Maine we can ditch and can always move the social icons to appear at the bottom on the viz.

mwernimont:
We aren't. It'll be fixed with future PR's of functionality for the mobile UI.

lindsaycarr:
To answer your question about buttons, I think we could shorten the drop downs a bit and make sure everything is visible. I think they might fit, but we won't be able to see the text. We know we want them on there somewhere, so we can get them up and then play around with the layout after.

lindsaycarr:
I'll let @aappling-usgs decide if we should close this issue and make a new issue for the problem mentioned here https://github.com/USGS-VIZLAB/water-use-15/pull/177#issuecomment-380883461, or if we should just keep this one open until it is addressed.

aappling-usgs:
As noted in #177, I believe that issue is captured in #160.

Issue #89, opened 2018-03-26
Prototype button options (S)


aappling-usgs:
Create a mock-up for category & water use volume legend layout for:
- national view in desktop
- national view in mobile
- state view in desktop
- state view in mobile

Initial discussion of options at https://docs.google.com/document/d/17Q0Ei2hxmlmFHgmSWZ2hxtzhKxnJJGKWfE5BqNl4LOY

aappling-usgs:
I like the overlay on desktop zoom. Add a touch more buffer (larger overlay rectangle)?

Also add a bit of vertical buffer between mobile map and mobile buttons?

I'm not sure I love the one longer bar. It's OK, but...what about making the grey bars shorter, so Thermoelectric has the same grey space on either side of the word? Or making them all long enough to always accommodate a number without growing? Or what about centering everything in mobile - text within bars, title, and bars (where one could be longer than the others but symmetrically left and right)?

Could we do smaller buttons in mobile? And adjust them a bit to the left, or else center them if we use all same-size bars?

Maybe drop 2015 from the button title, just keep it in the vizzy title.

Move the `(millions of gallons per day)` to be a subtitle to the legend title? I think people would notice & use that info more quickly if it were a subtitle.

Interesting idea to put the watermark by the footnote in mobile. Would moving the footnote to the subtitle mess that up aesthetically? I think it could also reasonably still live on the map, but I prefer what you've proposed (below the buttons).

What technology did you use to prototype these?

jread-usgs:
@aappling-usgs I used powerpoint to mock these up and would be happy to share the file if you would prefer that.

As for the bar lengths and details related to it, I was thinking we could wait on seeing how this would look w/ data, which will help determine necessary bar length. 

I was also thinking this category legend could take the place of the tooltips. When you mouse on the county the value and category title would change to update the information to the county being moused over.

aappling-usgs:
I like the legend = tooltip idea! Trick will be to make it likely that the user will be viewing both the map and the legend when they start playing around, which already seems very likely with the arrangements you've proposed (assuming we include the buttons in the auto-scaling in desktop mode). We might also need to make the change a bit flashy to help users notice (make the button color flare or size expand for 0.5 seconds or something), but I think it's safe to start with the expectation that changing the text will be enough to draw attention.

mwernimont:
[...]

PR #172, opened 2018-04-10
dropdown selector functionality - part way there

lindsaycarr:
Working towards #140. This is not totally perfect (there are some resetting behaviors that still need work). 

What this does:
* Populates the states into the view drop down on load.
* When the view is changed to a state, the counties are added to the county drop down and the zoom goes to that state.
* When the view is changed back to `United States - Water Use`, the view is changed back to national.
* When a county is selected in the drop, the county and appropriate circle are highlighted.
* When the view changes, the highlighted county/circle are reset.
* The "zoom out" button does not appear on load. 
* The zoom out button is hidden when the view selector changes to national view and appears when view selector changes to a state.

What this still needs (hoping all but maybe the first one can just come in a different PR):
1. Remove current counties when view changes rather than just appending (tried but can't get that darn ENTER-UPDATE-EXIT stuff down). 
1. Reset what is shown in the county dropdown when you switch states, should go back to `--Select County --`.
1. Hook up to the solution for showing values in the category selector (#139)

Things we could do:
1. Sleeker zoom out button & better way to indicate it is active or inactive (maybe just greyed out)
1. Some way to skip all of it if we are on mobile?
1. On mobile only, someone can click a state and zoom to that view, then click a county and highlight/get details. They need to use "zoom out" to get back to everything else?

dropdowns


js/dropdown_selector.js
```diff
@@ -0,0 +1,116 @@
...
+          var selectedView = menu.options[menu.selectedIndex].value;
+          updateView(selectedView); // letting analytics fire every time for now
+          
+          // reset county highlights
+          unhighlightCounty();
+          unhighlightCircle();
+          
+          // filter to correct county data then update dropdowns
+          var stateCountyData = countyCentroids
+            .filter(function(d) { return d.STATE_ABBV === selectedView; });
```
mwernimont: I was so close.

viz.yaml
```diff
@@ -281,6 +282,10 @@ publish:
     id: map_counties
     location: "js/counties.js"
     mimetype: application/javascript
+  -
+    id: dropdown_selector
+    location: "js/dropdown_selector.js"
```
mwernimont: Good idea to split this functionality up.  Was torn between doing that or just adding it to the map_utils, as it affects the map.  Probably smarter to keep it separated.

layout/templates/mobileMapUI.mustache
```diff
@@ -5,7 +5,7 @@
         {{{selectArrows}}}
       </div>
       <select class="view-select">
-        <option value="United States">United States - Water Use</option>
+        <option value="USA">United States - Water Use</option>
```
aappling-usgs: It doesn't make sense to me to have ` - Water Use` for United States only, not for states. I could see either adding ` - Water Use` to every state name or leaving it off of all of them. I'd advocate for leaving it off - I think that info is better suited for the header (where it already is) and/or the title or subtitle of the legend.

jread-usgs:
Looking cool!

Some of your comments are "on mobile" like "Some way to skip all of it if we are on mobile?", but isn't this PR for mobile only? a la #89 ?

js/build_map.js
```diff
@@ -129,6 +129,11 @@ function fillMap() {
   addCircles(circlesPaths);*/
   updateCircles(activeCategory);
   
+  // manipulate dropdowns
+  // need to figure out how to do this for only mobile views
```
aappling-usgs: I'm working on a switch between buttons on the left and buttons on the bottom, triggered by a resize and (so far, at least) whether the window width is < or > than 425. We could set a global variable for `mobileMode` in there and then tie this line to that variable, once my PR is in.

lindsaycarr:
@jread-usgs well that's just the issue I am having. All of my code will still affect the dropdown divs, but the only way you can actually see them is on mobile thanks to https://github.com/USGS-VIZLAB/water-use-15/blob/master/layout/css/custom.css#L193-L195. I'm not sure how to detect if we are in a mobile-sized window within the d3 code to use that to skip these functions.

js/styles.js
```diff
@@ -73,13 +73,13 @@ function resetState() {
...
-  d3.select(selection)
+  selection
     .classed("highlighted-county", true);
 }
 
 // on mouseout
-function unhighlightCounty(selection) {
-  d3.select(selection)
+function unhighlightCounty() {
+  d3.selectAll('.county')
```
aappling-usgs: This looks more computationally intensive than selecting and editing a single element. Is it possible to keep the selection in there, or maybe to call `d3.select('.highlighted-county')` here instead of selecting all the counties?

layout/templates/mobileMapUI.mustache
```diff
@@ -18,6 +18,6 @@
     </div>
   </div>
   <div class="zoom-out-button-container">
-    <button class="zoom-out-button">Zoom Out</button>
+    <button class="zoom-out-button hidden">Zoom Out</button>
```
aappling-usgs: Is it too early to also work on making the state-level selector expand and shrink when the zoom out button is removed or added?

js/dropdown_selector.js
```diff
@@ -0,0 +1,116 @@
...
+  
+  // add new options
+  countyOptions
+    .enter()
+    .append("option")
+      .property("value", function(d) { return d.GEOID; })
+      .text(function(d) { return d.COUNTY; });
+  
+  // remove old options -- NOT WORKING RIGHT NOW
+  // countyOptions.exit().remove();
```
aappling-usgs: Hmm, this looks reasonable enough to me. What error/behavior are you seeing?

mwernimont:
https://www.w3schools.com/howto/howto_js_media_queries.asp might be helpful?

aappling-usgs:
`how to detect if we are in a mobile-sized window within the d3 code to use that to skip these functions` - see my comment above about the function i'm working on. let's wait on that until i can get my PR in (about legends/category selects/buttons), because I think I can help.

In my PR I'm currently using window width as the way to decide whether to show buttons on the left or the bottom. I thought about looking specifically for mobile vs desktop but then figured we'd want the buttons to move when the screen is narrow regardless of whether it's mobile or desktop. Open to opinions on that, and on whether we should use different rules for legend placement vs dropdowns. Either way, I think a call to `window.matchMedia` within the `resize` function I'm working on would be fine to include.

js/styles.js
lindsaycarr: I couldn't figure out how to pass the previous county here (https://github.com/USGS-VIZLAB/water-use-15/pull/172/files#diff-5fd12accec11a9db4351f064170a4fd2R87) without setting some type of global variable, but yes - selecting just that one should work!

layout/templates/mobileMapUI.mustache
lindsaycarr: I think the resizing bit will be the difficult thing

mwernimont: I disagree, we just have to hide the div that the zoom-out button is in and flex-box will take over.

lindsaycarr: shouldn't that be happening then? I use `display: none` for the zoom-out box but the dropdowns stay the same size. Is there a different way I should be hiding them?

mwernimont: @lindsaycarr sorry was editing my post above, check it out now

lindsaycarr:
OK, worked on some of the suggested things. Here's what it's like to hide and resize when at national vs state views. Personally, the zooming into a state view seems fine, but zooming out seems choppy and weird to just have those things appear. Wondering how we make it smoother.

dropdowns-hide

aappling-usgs:
One more thing - could you sort the county names alphabetically?

PR #173, opened 2018-04-10
adding text (some still placeholder)

jread-usgs:


lindsaycarr:
Brings us that much closer to the real deal, so 👍

PR #176, opened 2018-04-11
appropriately updates counties

lindsaycarr:
When using the drop down to change states, the county list appropriately updates. The problem is that now the only single existing option `--Select County--` gets updated/overwritten, but I actually want to keep it as the first one. In D3, is there a way to update everything except for one specific one?

### DIFF
--- a/js/dropdown_selector.js
+++ b/js/dropdown_selector.js
@@ -108,25 +108,28 @@ function updateCountySelectorOptions(countyData) {
   
   // bind data to options in county dropdown menu
   var countyOptions = countyMenu
-        .selectAll("option")
-        .data(countyData);
-  
-  //update existing options with data
-  countyOptions
-    .property("value", function(d) { return d.GEOID; })
-    .text(function(d) { return d.COUNTY; });
+        .selectAll("option");
   
   // add new options
   countyOptions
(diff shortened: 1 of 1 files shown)


lindsaycarr:
[This video](https://www.youtube.com/watch?v=OZXYk_bgQGQ) helped me understand enter, exit, update more but it still isn't working how I think it should here.

```
DOM elements < data elements (ENTER, selection is DOM element placeholders to additional data)
DOM elements > data elements (EXIT, selection is DOM elements without data match)
DOM elements = data elements (UPDATE)
```

mwernimont:
This is a wild guess, but could you do a filter on the --Select County--?  So D3 doesn't remove it.  I'm sure Alison has a better idea though.

lindsaycarr:
the issue is that it doesn't have any data associated with it, so I'm not sure how to do the filter

lindsaycarr:
OK, new commits correctly update counties, keep "Select County" as top option, and reset to "Select County" when views are changed. Ready to merge @aappling-usgs and @mwernimont !

county_selector

aappling-usgs:
Looking great, @lindsaycarr and @mwernimont !

Issue #160, opened 2018-04-09
Revisit need for mouseover delays


aappling-usgs:
Could prevent accidental hovering (e.g., when you're really just moving your mouse to another part of the screen). Alternatively, if transitions or quick, could be better to have immediate responsiveness after all. 

Spinoff from #126

ldecicco-USGS:
So we're hoping this issue will fix the jumping up to the selected target. Increasing the delay could help, and other option might be to add an invisible box that eliminates the gaps between the colored boxes.

aappling-usgs:
I'm going to unassign myself because I won't get to this today or tomorrow, but I'm happy to reassign myself if it still needs doing next week.

Issue #127, opened 2018-04-04
Use namespacing for globals


aappling-usgs:
Javascript best practice is to use namespacing rather than globals (recommended by Mary at https://github.com/USGS-VIZLAB/water-use-15/pull/125#discussion_r179135637).

Namespacing appears to be easier than I thought - it can simply mean defining a single global (an object literal, i.e., a curly-braced object with named elements, e.g., `myObject = { a : 1, b : "hello" }`) and adding any desired global information to that single global object. (https://addyosmani.com/blog/essential-js-namespacing/#beginners, http://www.dyn-web.com/tutorials/object-literal/)

I didn't understand why people would call this namespacing instead of "only using one global variable", but Andy pointed out that the dot notation works for both packages and objects (e.g., `d3.scale`, `myVizzyNamespace.data`), so why not use it as a local namespace as well? Seems plenty practical as a way to lower the odds of mistaking a global for a local or having an outside package clobber one of our variables.

Immediately-invoked Function Expressions (IIFE)s can be used to generate all or part of a namespace global variable, but I don't think we'll have a need in this particular vizzy. See https://addyosmani.com/blog/essential-js-namespacing/#beginners for that idea and other ways to get fancy if and when we need to.

Our vizzy currently has a number of global variables - state and county boundary datasets, flags for whether circles have been created or not, etc. - that should use namespacing.

jread-usgs:
this was done already, right? `waterUseViz.stuff`?

ldecicco-USGS:
I'd vote we close this issue. We'll do better with the next one, but we got the major ones.

lindsaycarr:
Well, we have a bunch of things in the backlog (this included). I'm not sure we are going to close them. I think we will just leave them sitting in that swim lane. If it stresses you out to see so many open issues (it does for me), never look at the Issues view, and go to our project view instead (https://github.com/USGS-VIZLAB/water-use-15/projects/1). This is really where we've been tracking what needs to happen.

ldecicco-USGS:
good point, I'm just doing a sweep of what I think I could help with. I could do this...but I just don't see it worth my time right now.

PR #177, opened 2018-04-11
Button revamp

aappling-usgs:
Don't merge yet - I still have some tweaking to do, just want others who are changing a lot of code to see what I'm doing.

180410-buttons1


jread-usgs:
We’re getting close! :tada:

js/build_map.js
```diff
@@ -1,14 +1,43 @@
-// Width and height
-var chart_width     =   1000;
-var chart_height    =   700;
-
-// define categories
+// Define one big global to eventually rule them all.
```
lindsaycarr: 👑 I like this idea! Still let's us have globals, but less of a chance for having namespace var name conflicts and such.

js/build_map.js
```diff
@@ -1,14 +1,43 @@
...
+    svg: {
+      width: null,
+      height: null
+    },
+    watermark: {
+      width: null,
+      height: null
+    }
+  },
+  elements: {
```
lindsaycarr: What is the role of these entries?

js/build_map.js
```diff
@@ -134,32 +163,70 @@ function fillMap() {
...
 }
 
-function addButtons() {
-  var buttonContainer = d3.select('.svg-container')
-    .append('div')
-    .attr('id', 'button-container');
-    
-  buttonContainer.append('div')
-    .classed('select-arrowbox', true);
+function resize() {
```
lindsaycarr: ❤️ this function

lindsaycarr: I'm kind of thinking it needs a standalone file though.

js/build_map.js
```diff
@@ -134,32 +163,70 @@ function fillMap() {
...
+  
+  // Decide whether we're in mobile or desktop mode. Currently doing this by window width, but we could look to
+  // https://www.w3schools.com/howto/howto_js_media_queries.asp for more device-specific solutions
+  if(window.innerWidth > 425) { // sufficiently wide desktop windows
+    waterUseViz.mode = 'desktop';
```
lindsaycarr: does `mode` need to exist in the `waterUseViz` global variable to start? It currently does not.

js/build_map.js
```diff
@@ -134,32 +163,70 @@ function fillMap() {
...
+    // svg is [map]
+    //        [buttons]
+    waterUseViz.dims.svg.width = waterUseViz.dims.map.width;
+    waterUseViz.dims.svg.height = waterUseViz.dims.buttonBox.y0 + waterUseViz.dims.buttonBox.height;
+    // watermark is at bottom right
+    waterUseViz.dims.watermark.x0 = waterUseViz.dims.svg.width * 0.85;
+    waterUseViz.dims.watermark.y0 = waterUseViz.dims.svg.height * 0.95;
+  }
+  
+  // Apply the changes to the svg, map, map background, and watermark
```
lindsaycarr: So are these things global vars? Should they either be in `waterUseViz` or passed into this function?

js/build_map.js
```diff
@@ -134,32 +163,70 @@ function fillMap() {
...
+  // Apply the changes to the button elements
+  waterUseViz.elements.buttonBox
+    .attr('transform', 'translate(' + 0 + ', ' + waterUseViz.dims.buttonBox.y0 + ')');
+  waterUseViz.elements.buttonBox.select('#button-background')  
+    .attr('width', waterUseViz.dims.buttonBox.width);
+  waterUseViz.elements.buttonBox.selectAll('.button .category-label')
+    .attr('x', waterUseViz.dims.buttonBox.width * 0.05); // nudge a little over from the rectangle's left edge
+  waterUseViz.elements.buttonBox.selectAll('.button .category-amount')
+    .attr('x', waterUseViz.dims.buttonBox.width * 0.9); // nudge a little over from the rectangle's left edge
+  updateButtons(activeCategory); // updates the button rectangle widths
```
lindsaycarr: `activeCategory` would also be one of those global things, right?

js/buttons.js
```diff
@@ -0,0 +1,103 @@
+function addButtons() {
+  var buttonBox = waterUseViz.elements.buttonBox = svg.append('g')
```
lindsaycarr: woah - what's the result of having the two equals? Do `buttonBox` and `waterUseViz.elements.buttonBox` become the d3 group selection?

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+    .attr('id', 'button-container');
+  
+  buttonBox.append('rect')
+    .attr('id', 'button-background')
+    .attr('height', waterUseViz.dims.buttonBox.heightDesktop);
```
lindsaycarr: should this be just the generic height as in `waterUseViz.dims.buttonBox.height`? That var adjusts based on the window size where `heightDesktop` is static, right?

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+  
+  var buttonY = d3.scaleBand()
+    .domain(['title'].concat(categories))
+    .range([0, waterUseViz.dims.buttonBox.heightDesktop])
```
lindsaycarr: same comment about `waterUseViz.dims.buttonBox.height` vs `waterUseViz.dims.buttonBox.heightDesktop`

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+    .paddingInner(0.1);
+  
+  // one group for each button
+  var buttons = buttonBox.selectAll('rect.button')
+    .data(categories)
+    .enter()
+    .append('g')
+    .classed('button', true)
+    .attr('transform', function(d) {
```
lindsaycarr: should we add an `id` for each button group that is the category name?

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+    })
+    .attr('y', function(d) {
+      return buttonY.bandwidth()/2;
+    });
+
+  // button category use values
+  buttons.append('text')
+    .classed('category-amount', true)
+    .text(function(d){
+      return 'XX';
```
lindsaycarr: this will be the tricky part for today I think 😉

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+    })
+    .attr('width', function(d) {
+      if(d === category) {
+        return waterUseViz.dims.buttonBox.width * 0.95;
+      } else {
+        return waterUseViz.dims.buttonBox.width * 0.7;
+      }
+    });
+  
+  waterUseViz.elements.buttonBox.selectAll('.button .category-label')
```
lindsaycarr: what if we put these styles into a class called `button-selected` or something, and the if else was in classed here
```
.classed("selected", function(d) {
     if(d === category) {
          return true;
     } else {
          return false;
     }
});
```

js/buttons.js
```diff
@@ -0,0 +1,103 @@
...
+  
+}
+
+function updateButtons(category) {
+  waterUseViz.elements.buttonBox.selectAll('.button rect')
+    .style('fill', function(d) {
+      if(d === category) {
+        return categoryToColor(d);
+      } else {
+        return '#c6c6c6';
```
lindsaycarr: I think if we are defaulting them to being completely greyed out here, then we need some type of legend elsewhere for the pies below.

lindsaycarr: Although just after I wrote that, I realized that we have the national pie chart with labels. Do you think that would be enough?

js/build_map.js
aappling-usgs: good idea. it got pretty big pretty quick.

js/build_map.js
aappling-usgs: good! I believe this is a pretty standard way of doing namespacing in javascript, so it gets us a start on #127

js/build_map.js
aappling-usgs: i picture this eventually containing all of the DOM elements that we want to keep as globals - `svg` and `map` for sure, probably `watermark` and `mapBackground`. For now I'm just adding the one DOM element that I'm currently editing (`buttonBox`) so I don't change too much code at once.

js/build_map.js
aappling-usgs: It does not need to exist initially - this line will add it - but I like the idea of outlining the full global object all in one place, so I'll initialize it up top as `null`.

js/build_map.js
aappling-usgs: they are globals, and i'd like to get them in waterUseViz eventually, but I'm trying not to change too too much code in this PR, and #127 is a separate (backlogged) issue. so i'm really just trying to lay the groundwork so that someone can complete the transition to using `waterUseViz` in the future.

js/buttons.js
aappling-usgs: `waterUseViz.dims.buttonBox.height` isn't yet initialized at this point, which is why I had to use the default.

js/buttons.js
aappling-usgs: same response =)

js/buttons.js
aappling-usgs: sure, could do that. i don't think we have a direct need for it, but it probably wouldn't hurt

js/buttons.js
aappling-usgs: maybe so =). and a separate PR.

js/buttons.js
aappling-usgs: i played with some options for that. one challenge is that the selected button has a different color from any other selected button, so a `selected` class for css would have to omit color, and width is calculated (and i'd like it to be transitioned) and so would also need to be a d3 thing. i also tried defining the default color in one css definition and the specific colors in other css/d3 locations, but i couldn't get the priorities right in my first effort. i'll take one more shot at using more css for this, at least for a subset of the properties.

js/buttons.js
aappling-usgs: Hmm, yes and yes. I'm following https://github.com/USGS-VIZLAB/water-use-15/issues/89#issuecomment-378579693, which I do think looks nice.

js/build_map.js
aappling-usgs: yes, and see above - saving that for issue #127.

js/buttons.js
aappling-usgs: You can do this in R, too. Yes, the elements on the left and in the middle get set to the rightmost value. I'm just using it here so I don't have to spell out `waterUseViz.elements.buttonBox` everywhere in this function.

lindsaycarr:
@aappling-usgs we're leaving this PR alone until you say otherwise, right?

lindsaycarr:
I'd love to have some of this resizing stuff soon though because I'm going to work on the changes to mobile click behavior.

aappling-usgs:
yes, please don't merge yet. i haven't tried bringing in the recently merged PRs yet. i swear i'm trying but i've only found 1 hour for this all day so far. i'll try to push through without getting too picky about style tweaks.

aappling-usgs:
OK, I think I'm ready for final review now. Added title and subtitle, positioned the watermark better, and futzed with sizes and alignments so it looks decent. The merge with upstream was surprisingly easy.

Here's the transition from desktop to narrow desktop. note that zoomed-in non-focal states share space with the buttons, which is consistent with JR's prototypes and I think I like it (sure hope so, because otherwise I did a lot of stuff in svg that could have been better achieved with divs!):
180411-buttons-desktop

Looks fine on multiple mobile models:
180411-buttons-mobile-models

Here's the behavior on mobile:
180411-buttons-mobile-zoom

aappling-usgs:
Fixes #139. I spent 6 hours on this.

lindsaycarr:
I'm going to merge, so we can get this functionality in there but there is some weirdness with switching categories, where it jumps back up to Total. It might be because of a delay somewhere? Not sure, but here's what I see (I am on Chrome):

weird_skipping

aappling-usgs:
I think that issue about jumping is actually due to a _lack_ of delay and can be solved by resolving #160 - adding a delay before we respond to mouse over or mouseout on those buttons. I think what's happening is that when your mouse goes over some white space, the display reverts to the currently active category (Total in your example), which is what we want it to do when, for example, you stop hovering on buttons and move your mouse over to the map. It's just doing the category switches too responsively.

jread-usgs:
I think that is expected behavior, since the category goes back to the active one when the mouse is off the rectangle. We probably want to modify that, yes, but I don't think there is an error involved.

aappling-usgs:
Jinx =)

Issue #135, opened 2018-04-05
Comprehensive style unification sweep


aappling-usgs:
Do this just before the submission. Includes revisiting the state borders, colors, zoom styling, etc.

aappling-usgs:
* Lcarr: personally think the background colors, lines need some work. White background + dark black outlines for state bounds look somewhat outdated
* Make sure the active category button is visibly selected (e.g., darker or more colorful) relative to the others

jread-usgs:
a quick note (maybe this goes to style sweep?) but the title doesn't change when the category changes on hover. I assume we don't need to worry about this because of the category selector change, but something to keep in mind

aappling-usgs:
We're deleting the title, right? I was partway through creating an issue to do that.

jread-usgs:
yes, that is my understanding

jread-usgs:
Note counties in AK have things like "Census" in the name. I assume we'll want to modify that.

jread-usgs:
- [ ] The zoom in on PR clips parts of the VI. We'll need to address it, but I don't think we should tinker until the new category switchers are in place. 
 - [ ] Our user experience in the NE needs some work, but I am not sure I would call it a priority this week. It is really hard to see what county you are getting values from when zoomed in (for example) to New Jersey and looking at total or thermo. I don't know if there are simpler solutions, but my suggestion later on would be to use a `use` path for the county to create a copy of the county outline (no fill) that sits on top of everything and has a thin white-ish stroke. 
 - [ ] map outlines and fills
 - [ ] county boundaries for non-active state look exaggerated when scaling
 - [ ] main blue color for total is too purple and we want to get this color right, since it is the one that everyone will see 
 - [ ] layout needs some work and perhaps some separation between the different concepts. Current view has things kind of floating out there (pie, pie map, placeholder rank-'em states). Hesitant to put too much energy _right_ now since we are still adding and refining content
 - [ ] we may want to revisit the zoom-to rules when the category selectors are implemented, since on desktop, these will take up the left side of the zoom box. 
 - [ ] rounding/styling water use values shown in the category legend (e.g., 2.00000021)
 - [ ] entering the viz w/ a state selected doesn't give you the same style properties as zooming to the state from a national starting point **not a priority for this week**

aappling-usgs:
> Note counties in AK have things like "Census" in the name. I assume we'll want to modify that.
What would you use instead of Census Area? It looks like a legit description of a county equivalent to me.

[...]

PR #178, opened 2018-04-11
parts of #135

jread-usgs:
from #135:
~map outlines and fills~
image
vs
image

~county boundaries for non-active state look exaggerated when scaling~

~main blue color for total is too purple and we want to get this color right, since it is the one that everyone will see~
is now:
image

I also changed the county mouseover style at the national level:
image
vs
image
and when zoomed in

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -134,12 +134,14 @@ function updateView(newView, fireAnalytics) {
     backgroundState(otherstates, scale = zoom_scale);
     foregroundState(thisstate, scale = zoom_scale);
     
-    statecounties
-      .transition()
-      .duration(500) 
-      .style("stroke-width",  0.75/zoom_scale); // make all counties have scaled stroke-width
+    
   }
-
+  var allcounties = d3.selectAll('.county');
+  
+  allcounties.transition()
+    .duration(500) 
+    .style("stroke-width",  1/zoom_scale); // make all counties have scaled stroke-width
+  
   // apply the transform (i.e., actually zoom in or out)
   map.transition()
     .duration(750)
--- a/js/styles.js
+++ b/js/styles.js
@@ -10,7 +10,7 @@ function categoryToName(category) {
 }
 
 function categoryToColor(category) {
-  if (category === "total") { return "rgba(38, 120, 178, 0.8)"; }
+  if (category === "total") { return "rgba(38, 140, 178, 0.8)"; }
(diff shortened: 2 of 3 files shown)


js/map_utils.js
```diff
@@ -134,12 +134,14 @@ function updateView(newView, fireAnalytics) {
...
     foregroundState(thisstate, scale = zoom_scale);
     
-    statecounties
-      .transition()
-      .duration(500) 
-      .style("stroke-width",  0.75/zoom_scale); // make all counties have scaled stroke-width
+    
   }
-
+  var allcounties = d3.selectAll('.county');
```
lindsaycarr: @aappling-usgs has noted this to me before, but I think we would be unnecessarily selecting and changing all 3224 counties when we only need to change the stroke of the current state's counties. This might affect performance.

lindsaycarr:
Yay! 🎉

wdwatkins:
Lighthouse was dinging the total color for not enough contrast with the background, is this fixing that too?

js/map_utils.js
jread-usgs: I did this on purpose - the alternative is that the other state county doesn't start scaled the right way when panned to. To see this check dev and zoom to NJ, then click NY and see the borders start huge.

jread-usgs:
David, I didn't see that, but I assume it will at least help, since the background has been de-emphasized.

js/map_utils.js
lindsaycarr: Hmmm OK. I see now. I wonder if we can just chop off the transition duration instead?

jread-usgs: I tried it several ways, including chopping the duration to 0 (side note, if you do that, don't you want to use D3 to set the style `.style("stroke-width"..)` and not use a transition at all?) and comparing performance of the whole US vs just the counties of the highlighted states. 

I couldn't see an impact of either of the approaches on performance, but cutting off the transition does leave the somewhat undesired artifact of having the county you clicked on to zoom out immediately get thick while other things smooth through. 

I looked on chrome (and throttled the CPU) and tried this on FF visually. But all of these are on mac and I sense windows may be different and if we are worried about that we should try it out. It ends up being a matter of using `allcounties` or `statecounties` and whether the transition only happens within the `if(activeView !== 'USA')` or regardless in case someone wants to give it a spin.

lindsaycarr: if you didn't notice any significant performance issues, then I'm fine with how you've implemented it! It does look smoother overall.

PR #182, opened 2018-04-12
Rank em

jread-usgs:
This is a start on the interactive fig for figure 4, where users can try ranking three states after we give them values for the remaining states/territories. 

I moved the states around in the map visual (fig 1) by creating `<defs>` and turning the references to the states from `<path>` to `<use>`. I did this so I could re-use the same paths in this figure w/o having to read in the data and retroject it. I think this is a really efficient way to do things, but if there are performance issues w/ `<use>`, we should be aware of them.

I made some fake data for this and am assuming we'll create json in a processing task to do this. I am not up to speed with the various ways to read in external data in d3, so I skipped over that and hard-coded example data. 

I assume there are better ways to filter selections that I am doing (e.g., putting a bunch of `if else` logic in that depends on classes). 

I also assume there is an easy way to create the movable states based on the `bardata` array's value for `open` instead of having a separate array of movable state names. 

What I haven't done yet: 
 - it doesn't tell you when your guess is wrong
 - add real data
 - add mouseovers so you can get state names or stuff like that - tabling that for now
 - figured out how to deal with the mix of classes and style properties that are in js or css. I am really enjoying d3, but finding it challenging to get the proper way to deal with js vs css style stuff into my head.
 - clean up details
 - gotten peer review on this approach ;) 

what it looks like now: 
ezgif com-video-to-gif

You can pick up a state and "drop" it - will go back to its original place if you aren't hovering over a bar or it will become a bar if you are on an open one. Supporting users being able to drop into a new place and then pick it back up seemed to not be worth the complexity to me.


js/rank_states_utils.js
```diff
@@ -3,3 +3,142 @@ var svgStates = d3.select("#rank-states-interactive")
...
+
+var stateMap = svgStates.append('g')
+  .attr('id','ranked-states-map')
+  .attr('transform',"scale(0.4)");
+
+stateMap.append('g')
+  .attr('id','ranked-states-moved');
+
+stateMap.append('g')
+  .attr('id','ranked-states-dragable');
```
lindsaycarr: just to be consistent since the classes spell this right, might want to add another `g` 😄

js/rank_states_utils.js
```diff
@@ -3,3 +3,142 @@ var svgStates = d3.select("#rank-states-interactive")
...
+var dragStates = ["ID","OK","MI"];
+
+// add states
+  svgStates.select('#ranked-states-moved')
+    .selectAll( 'use' )
+    .data(movedStates)
+    .enter()
+    .append('use')
+    .style('stroke-dasharray',"10, 10")
+    .classed('moved-state', true)
```
lindsaycarr: calling these `moved-state` is a little confusing. It sounds like the ones that are going to be moved by the user. Maybe it should be `static-state` and `draggable-state`

js/rank_states_utils.js
```diff
@@ -3,3 +3,142 @@ var svgStates = d3.select("#rank-states-interactive")
...
+        null;
+      }
+    })
+    .on('mouseover', function(d){
+      if (d.open){
+        d3.select(this)
+          .style("stroke-dasharray",null)
+          .classed('chosen-rank-bar',true)
+          .style("fill", categoryToColor("total"));
+      } else {
```
lindsaycarr: you don't need to say `else { null; }`. Same with mouseout function below

js/rank_states_utils.js
```diff
@@ -3,3 +3,142 @@ var svgStates = d3.select("#rank-states-interactive")
...
+        d3.select(this)
+          .style("stroke-dasharray","4, 2")
+          .classed('chosen-rank-bar',false);
+        } else {
+          null;
+        }
+      });
+  
+    function dragging(d) {
+      d3.select(this).attr("transform", "translate(" + (d.x = d3.event.x) + "," + (d.y = d3.event.y) + ")");
```
lindsaycarr: I don't think you need `d.x =` and `d.y =` since you are specifying a translate value

js/rank_states_utils.js
```diff
@@ -3,3 +3,142 @@ var svgStates = d3.select("#rank-states-interactive")
...
+          .on('mouseover', null)
+          .on('mouseout', null)
+          .attr('class','closed-rank-bar');
+        d3.select(barchoice.node().parentNode).select('text')
+          .classed('open-bar-name',false);
+        d3.select(this)
+          .transition().duration(600)
+            .attr('opacity', 0); 
+      }
+      d.x = 0;
```
lindsaycarr: these two lines `d.x` and `d.y` don't seem to be doing anything. I think they can be deleted.

js/rank_states_utils.js
jread-usgs: Gotcha - these d.x and d.y parts are holdovers from when I was letting the state be dropped and picked back up, and we needed to capture the origin of the event. It didn't work very well anyhow..

js/map_utils.js
```diff
@@ -8,18 +8,33 @@ var zoom_scale;
...
+    .enter()
+    .append('use')
     .classed('state', true)
     .attr('id', function(d) {
       return d.properties.STATE_ABBV;
     })
-    .attr('d', buildPath);
+    .attr('xlink:href', function(d) {
+      return '#'+d.properties.STATE_ABBV+'-pattern';
+    });
```
aappling-usgs: note that state bounds are not entirely static once they're loaded - they switch between low-resolution (for national view) and higher resolution (for state view). The `addStates` function defines the initialization of these boundaries to the low-resolution dataset, while the `updateStateBounds` function (called in `updateStates`, which is called in `updateView`, anytime you zoom anywhere) drives the switches between these two datasets: https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/states.js#L24-L29.
* To keep the first figure functioning well, the obvious solution to me is to modify the `xlink:href` instead of `data` and `d` within `updateStateBounds` (https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/states.js#L24-L29).
* This one change would make it so that figure 4 will sometimes show low-res and sometimes show high-res state boundaries, depending on the current zoom state of figure 1.
* If we'd like figure 4 to always show low-res state boundaries, I think we need two sets of polygon definitions in `state-bounds-defs` (or a new defs collection, `state-bounds-defs-hi-res`), and then updateStateBounds should switch which definitions are pointed to rather than building a new path. Making this change might actually give us a performance boost to figure 1 as a nice side effect, because we won't be reprojecting each time we switch between national and state views, though it's possible that switching which element is `use`d is more expensive than I think - I say let's just try it and try to notice whether it's a net speedup or slowdown for fig 1.
* If we do create two collections or subcollections of polygon definitions, let's continue to pay attention to when each of those collections gets created. The hi-res state bounds are neither downloaded nor added into the DOM until the first time you zoom in, which I think is a good thing for initial page load time, so I'd like to retain that. I think this means that the building of `state-bounds-defs-hi-res` (or whatever it's naemd) should occur within `updateStateData` right after the data are loaded, probably right here: https://github.com/USGS-VIZLAB/water-use-15/blob/d4fcf0182efbbf34cb3aa173d96d599f5c49095c/js/states.js#L13.

js/rank_states_utils.js
```diff
@@ -3,3 +3,134 @@ var svgStates = d3.select("#rank-states-interactive")
...
+      return d.wu;
+    })
+    .attr('width','20')
+    .style("stroke-dasharray","4, 2")
+    .classed('open-rank-bar', function(d){
+      return d.open;
+    })
+    .classed('closed-rank-bar', function(d){
+      return !d.open;
+    })
```
aappling-usgs: Is it necessary to have classes for both both open and closed bars? As an alternative, could one of these states be managed by a css definition for a shared selection (something like `#ranked-states-bars g rect`?), while the other state is managed by an overriding (more specific) css selector? It's probably not too computationally expensive to loop through the bars to assign classes twice (there are only 52 'states'), but we went through a similar exercise with `hidden` vs `shown` and were able to collapse it back to just `hidden` after some thought - seems like a generally more streamlined approach to me.

js/rank_states_utils.js
```diff
@@ -3,3 +3,134 @@ var svgStates = d3.select("#rank-states-interactive")
...
+          .on('mouseover', null)
+          .on('mouseout', null)
+          .attr('class','closed-rank-bar');
+        d3.select(barchoice.node().parentNode).select('text')
+          .classed('open-bar-name',false);
+        d3.select(this)
+          .transition().duration(600)
+            .attr('opacity', 0); 
+      }
+    }
```
aappling-usgs: This is cool =). I haven't played much with dragging in d3 and appreciate a chance to see an example!

js/map_utils.js
jread-usgs: Good stuff! I'm thankful for peer review, because I totally whiffed on that one. Changes coming. 

I think I will try putting both resolutions in `use` first and see how that goes. I will probably wait on any class changes because I think there will be a bunch more changes that could wipe that out - specifically, my solution to rely on hovering over the bar doesn't work on mobile.

jread-usgs:
I added both resolutions to the `defs`, but I am not sure this is a good solution for us. I think it makes performance a little worse, but it is hard to tell. One thing it does differently that I don't like is it blanks out the lower resolution borders during the first zoom (while loading in the high res into `<defs>`?) and that doesn't look great. Looks fine during any other zoom/pan events.

jread-usgs:
I fixed the zoom-in load issue by sticking the callback back into the right spot. I thought I was simplifying this, but realized I didn't know how those things work.

lindsaycarr:
So the way @jread-usgs just explained this to me: `use` and `defs` add a little bit of complication. He would only feel strongly about going this direction as opposed to just loading the spatial data and reprojecting IF we are planning to make the pie map an svg. 

Two thoughts I have: 
1) Personally, I think that even just avoiding loading and reprojecting once for the ranking figure is good. Especially since the complicated part of getting `use` into the code has already been done with this PR. 

2) I am leaning towards not making the pie map into an SVG. We abandoned the previous attempt at putting pies into the svg because it introduces a lot of extra nodes and affected performance. I also think that we then open the doors for even more work because we will need to optimize the behavior on mobile,too. I think that since we are over hours from the original budget, the hours we do have left can be used to make adjustments for the first figures mobile behavior & make the rank em figure work a little better. If we finish all of those other things and decide that spending extra hours on making the pie map into an SVG would immensely improve what we have and/or we are gaining some valuable new d3 knowledge, then I think we should consider it.

jread-usgs:
Thanks Lindsay - excellent overview. 

I think I would feel strongly that if we had three map svgs in the viz that we made sure we weren't going to end up with issues related to performance based on unnecessary DOM complexity. 

Having an image instead of SVG for the pie map does simplify things and I think one thing we'll want to consider going that route is that we take advantage of having different resolutions of that image available and maybe take advantage of some image formats for certain browsers (chrome) that can give us a lot more compression. That is a discussion for a different task/issue though.

js/states.js
```diff
@@ -4,27 +4,44 @@ function updateStates(newView) {
...
+  
   map.select('#state-bounds')
-    .selectAll( 'path.state' )
-    .data(stateBounds.features, function(d) {
-      return d.properties.STATE_ABBV;
-    })
-    .attr('d', buildPath);  
+    .selectAll( 'use' )
+    .attr('xlink:href', function(d) {
+      return '#' + d.properties.STATE_ABBV + '-' + resolution;
```
aappling-usgs: :+1:

Issue #180, opened 2018-04-11
Add links between clicks/taps and dropdown


aappling-usgs:
Currently most click/tap-based behaviors should but don't change what's visible in the drop down:
* zoom-taps to or from from national view
* pan-taps from state view to another state
These behaviors work (as specified in #162), they just don't change the dropdown.

aappling-usgs:
![image](https://user-images.githubusercontent.com/12039957/38649045-4ac4cc08-3da9-11e8-8781-ed1bc94b6db4.png)

lindsaycarr:
Just making a note - I think this issue should create helper functions for what is shown in the drop down. And the way to update what is currently selected in the dropdown is:

```
activeView = "HI";
d3.select(".view-select").property ("selected", activeView);
```

The `value` field for the option elements is set as the state abbreviation for the view selector and county GEOIDs for the county selector

lindsaycarr:
@aappling-usgs I think this is done with #184.

PR #184, opened 2018-04-13
Mobile click

### LINKED ISSUE #162: Adapt click behavior for mobile
I think we're looking for these behaviors:
- [ ] from national view, tap on state zooms to state
- [ ] within state view, tap on county (county in selected state) gives county info
- [ ] within state view, tap on adjacent state pans to that state
- [ ] within state view, taps in the ocean/canada/mexico should still zoom you out
- no hover or pinch behaviors

lindsaycarr:
This fixes #162 and #180. 

Mobile clicking behavior (in addition to dropdowns): when you're at national view, you can click a state to zoom in. While zoomed in, you can click adjacent states and be brought to them. Each time you go to a new state, county highlights are refreshed. On your zoomed in state, you can click the counties to highlight them.

Clicking a state or county changes and updates the drop down appropriately.

There's still a little weirdness with when the tooltips are there or go away, but we are planning to take them out, so I wasn't worried about it.

mobile_clicking

### DIFF
--- a/js/counties.js
+++ b/js/counties.js
@@ -119,7 +119,31 @@ function displayCountyBounds(error, activeCountyData) {
         unhighlightCircle();
         hideToolTip();
       })
-      .on('click', zoomToFromState);
+      .on('click', function(d) {
+        
+        // clicking a county on mobile has no affect on the 
+        // view unless it's the same county as last time
+        if(waterUseViz.mode === "mobile") {
+          
+          var prevCounty = waterUseViz.prevClickCounty;
+          var thisCountyID = d3.select(this).attr("id");
+          if(prevCounty === thisCountyID) {
+            
+            //only zoom out if you click on the same county 
+            zoomToFromState(d, d3.select(this));
+            
+            // hide on any zoom bc no county will be selected
+            unhighlightCounty();
+            unhighlightCircle();
+            hideToolTip(); 
+          }
+          // set prevClickCounty as global var for next click
+          waterUseViz.prevClickCounty = thisCountyID;
+        } else {
+          // desktop county clicks zoom in and out
+          zoomToFromState(d, d3.select(this));
+        }
+      });
     
     // update
     countyBounds
--- a/js/counties.js
+++ b/js/counties.js
@@ -110,7 +110,7 @@ function displayCountyBounds(error, activeCountyData) {
       .on("mouseover", function(d) {
         highlightCounty(d3.select(this)); 
         highlightCircle(d.properties, activeCategory);
-        showToolTip(d, activeCategory); 
+        showToolTip(d.properties, activeCategory); 
         // OK to use global var activeCategory which only changes on click 
         // because people won't be able to hover on tooltips at the same time as hovering buttons
       })
@@ -136,7 +136,9 @@ function displayCountyBounds(error, activeCountyData) {
             unhighlightCounty();
             unhighlightCircle();
             hideToolTip(); 
-          }
+          } else {
+            updateCountySelector(thisCountyID);
+          }          
           // set prevClickCounty as global var for next click
           waterUseViz.prevClickCounty = thisCountyID;
         } else {
--- a/js/counties.js
+++ b/js/counties.js
@@ -119,7 +119,7 @@ function displayCountyBounds(error, activeCountyData) {
         unhighlightCircle();
         hideToolTip();
       })
-      .on('click', function(d) {
+      .on('click', function(d,i,j) {
         
         // clicking a county on mobile has no affect on the 
         // view unless it's the same county as last time
@@ -130,7 +130,7 @@ function displayCountyBounds(error, activeCountyData) {
           if(prevCounty === thisCountyID) {
             
             //only zoom out if you click on the same county 
-            zoomToFromState(d, d3.select(this));
+            zoomToFromState(d,i,j, d3.select(this));
             
             // hide on any zoom bc no county will be selected
             unhighlightCounty();
@@ -143,7 +143,7 @@ function displayCountyBounds(error, activeCountyData) {
           waterUseViz.prevClickCounty = thisCountyID;
         } else {
           // desktop county clicks zoom in and out
-          zoomToFromState(d, d3.select(this));
+          zoomToFromState(d,i,j, d3.select(this));
         }
       });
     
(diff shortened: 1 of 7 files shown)


js/map_utils.js
```diff
@@ -7,6 +7,12 @@ var zoom_scale;
 // Create the state polygons
 function addStates(map, stateBounds) {
 
+  // changes pointer events depending on mobile or desktop
+  var clickClass = "state-click-off"; // default to desktop
+  if(waterUseViz.mode === "mobile") {
+    clickClass = "state-click-on";
```
aappling-usgs: i'm surprised that this issue required modifying state or county click behaviors at the national view level. isn't the perceived click/tap behavior in this view identical regardless of whether you're in desktop or mobile? in both cases, you click/tap within a state boundary and get zoomed in on that state. at least in desktop, we can hide from the user the implementation detail that it's actually a click on a county polygon, not a state polygon, that technically triggers the zoom.

on the other hand, maybe, this change in behavior could give us the option of not loading county boundaries at all until the national view is fully rendered and/or the user zooms in. but this PR doesn't take that shortcut, right? so if we want to do that, maybe we should introduce a new issue to optimize the loading?

js/map_utils.js
```diff
@@ -36,15 +44,21 @@ function addStates(map, stateBounds) {
...
 // on click
-function zoomToFromState(data) {
-
+function zoomToFromState(d, selection = null) {
+  
+  if(typeof selection === "number") { // because `i` is defaulted
+    // couldn't rely on this when I changed from `.on("click", zoomToFromState)`
+    // to `.on("click", function(d) { zoomToFromState(); })`
+    selection = d3.select(this);
+  }
```
aappling-usgs: so what happens if selection is null? then we proceed to call things like `selection.attr()`? seems like that would lead to errors, wouldn't it?

js/states.js
```diff
@@ -26,5 +26,6 @@ function updateStateBounds(error, stateBounds) {
     .data(stateBounds.features, function(d) {
       return d.properties.STATE_ABBV;
     })
-    .attr('d', buildPath);  
+    .attr('d', buildPath)
+    .on('click', zoomToFromState); //only clicks on mobile because of pointer-events
```
aappling-usgs: i don't think we should need this `.on('click')` call in here because all elements of `#state-bounds` already exist (were created in `addStates`) and so should already have this method. You'll also note that there's a minor merge conflict with jordan's PR here; I think you may be able to adopt his code for this function wholesale.

this reminds me - i was worried about moving too much code earlier, but this might be an opportune time to move the `addStates` function into states.js, if you wouldn't mind.

js/counties.js
```diff
@@ -119,7 +119,33 @@ function displayCountyBounds(error, activeCountyData) {
...
+        
+        // clicking a county on mobile has no affect on the 
+        // view unless it's the same county as last time
+        if(waterUseViz.mode === "mobile") {
+          
+          var prevCounty = waterUseViz.prevClickCounty;
+          var thisCountyID = d3.select(this).attr("id");
+          if(prevCounty === thisCountyID) {
+            
+            //only zoom out if you click on the same county
```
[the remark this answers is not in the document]
lindsaycarr: So I don't think we have a choice to make tapping any county zoom you out. On mobile, you need to tap a county in order for it's data to display which was the major motivation for making this change. So, if you could click anywhere and it zoomed you out, you would never be able to actually get other county information. On desktop, you can just hover so that's not an issue. 

On mobile, the dropdown and zoom out buttons are available for counties that are too small to tap. I honestly don't know if there is a better way to do the zooming and county select w/ info behavior on mobile unless we completely remove the ability to click on the map and force people to use the dropdown from the get-go.

js/map_utils.js
lindsaycarr: County clicks are different on mobile than desktop. On mobile, county clicks give you the water use value but on desktop you can hover and get that information. Therefore, to get the state-level water use data on mobile, you would need to be able to click to get that information but clicking a county from the national view would skip that piece.

js/map_utils.js
lindsaycarr: Well, it shouldn't ever be null. I need to remove the default, I think. If you use this function like `.on("click", zoomToFromState)`, d is automatically `d` and selection is automatically `i` as in `function(d,i)`. So selection should never be null. Otherwise, it is being use where you explicitly are passing d and selection in, see https://github.com/USGS-VIZLAB/water-use-15/pull/184/files/300afba47e390b96bf1c6fbd5857de8f8abed8c6#diff-dbac7df67f73b73347fbe15446a0ba2dR133

js/counties.js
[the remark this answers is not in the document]
aappling-usgs: I agree, it shouldn't be that you can tap anywhere to zoom out. I was thinking instead that your only option for zooming out would be to use the Zoom Out button. I think if that were the only option, then users would experience it as cheaper to try and try again to shift focus to the county they're trying to see (because they wouldn't ever get zoomed out accidentally when they were just trying to focus on the tiny county they wanted)

js/dropdown_selector.js
```diff
@@ -68,20 +65,38 @@ function updateViewSelectorOptions(view, stateBounds, countyCentroids) {
...
   updateCountySelectorDropdown(view);
 
 }
 
-function updateCountySelectorOptions(countyData) {
+function updateStateSelector(view) {
+  d3.select(".view-select")
+    .selectAll("option")
+    .property("selected", function(d) {
+      return d.properties.STATE_ABBV === view;
```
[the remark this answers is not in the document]
lindsaycarr: Huh, good point. I think it has magically been returning FALSE for everything, so it defaults to the first one which is USA. I will update to be very explicit about what's going on for `view = 'USA'`

js/states.js
lindsaycarr: on second thought, I'm not going to move it because there are some global variables at the top of `map_utils.js` that will make this slightly more complex

js/counties.js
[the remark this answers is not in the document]
lindsaycarr: Jordan talked with me a bit about this today and we think that scenario could be an edge case. I think it would be more frustrating as a user to be able to tap to zoom in, but not to zoom back out. I'd like to leave as-is for now and revisit once we make a push to beta so we can all play around on mobile and try to see what still feels clunky.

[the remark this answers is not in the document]
aappling-usgs: Works for me. Thanks for thinking & talking it through.

lindsaycarr:
Work from today was ~2 hrs. Those pesky enter/exit/updates.

js/map_utils.js
aappling-usgs: Created #188 to optimize mobile loading to take advantage of this PR.

Issue #183, opened 2018-04-12
state summarized water use values


lindsaycarr:
Data processing task in R to get a JSON file of the state water use totals (so summaries of the "total" category). This PR https://github.com/USGS-VIZLAB/water-use-15/pull/168/commits/a8c996053b000209f4bf8d6419fd695d050247fd did it for the entire nation by category, so it would be something similar. 

In this step, we should be able to declare an `open` column to indicate which states we want to have as the draggable ones. Here's the dummy data that was created for getting started on the rank states fig. This should output JSON similar in format:

```
[
  {"wu":80, "abrv":'OK', "open": true}, 
  {"wu":140, "abrv":'WI', "open":false}, 
  {"wu":180, "abrv":'MI', "open":true}, 
  {"wu":210, "abrv":'ID', "open":true}, 
  {"wu":260, "abrv":'TX', "open":false},
  {"wu":310, "abrv":'CA', "open":false}
]
```

Related to #115.

jread-usgs:
can we tack on the first part of #186 to this?
- [ ] read in that json data and use it to build the chart
if that is cool...

lindsaycarr:
I was trying to make sure that this task did not involve any knowledge of d3, so I'd rather not let the first of #186 bleed into this one. Any strong reason why it should?

jread-usgs:
No, I don't have a strong reason. Just seemed like a good way to bake in the acceptance criteria for the dataset

ldecicco-USGS:
OK, I was looking at the mini-PR instead of the big one. My understanding now is it's literally/hard-coded that we're "open"ing OK, MI, and ID...ie...figuring out the flow to go from our "process" -> json (to make `bardata`) -> D3. 

Unless I'm mistaken, there is no need for a full state summary json file?

lindsaycarr:
@aappling-usgs yep - that's what I would say!

lindsaycarr:
Closed by #191. Hours estimate @ldecicco-USGS ?

ldecicco-USGS:
2 hrs

PR #185, opened 2018-04-16
Rank em help for #183

jread-usgs:
@lindsaycarr

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -4,6 +4,11 @@
 var nationDims;
 var zoom_scale;
 
+function extractNames(stateBounds) {
+  for (var i = 0, len = stateBounds.features.length; i < len; i++) {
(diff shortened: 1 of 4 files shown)


js/map_utils.js
```diff
@@ -4,6 +4,11 @@
 var nationDims;
 var zoom_scale;
 
+function extractNames(stateBounds) {
+  for (var i = 0, len = stateBounds.features.length; i < len; i++) {
+    waterUseViz.stateAbrvs.push(stateBounds.features[i].properties.STATE_ABBV);
+  }
+}
```
lindsaycarr: there is a `forEach` that allows you to clean this up. I haven't had too much experience with it, but I played with it and you can do this:

```
stateBounds.features.forEach(function(d) {
     waterUseViz.stateAbrvs.push(d.properties.STATE_ABBV);
});
```

Issue #62, opened 2018-03-02
Load time performance


jread-usgs:
Putting this in for reference with our first dev deployment that was made off of [this commit](https://github.com/USGS-VIZLAB/water-use-15/commit/71998dc010e1b9a584860d6bc9654460ab840372)

From dev on a fresh browser (it caches, so the second time is faster), we have a 4.38s load time, and we don't see the map until that is complete. Seems to be fairly consistent

![image](https://user-images.githubusercontent.com/2349007/36911135-3f1482ce-1e08-11e8-8fc8-77e363e05dd6.png)

I think this is our baseline, and we can reference this issue in the future as we improve and learn what changes impact this load time.

aappling-usgs:
#25 should help with this. so could presenting a low-resolution set of county boundaries for the nation view, only displaying (and loading) states and not counties for the nation view, and/or combining the state and nation topo lines into a single file.

jread-usgs:
Just so we know what we are shooting for before release - an average time before users see a map of < 400ms. (Jim + Jordan's opinion)

jread-usgs:
Seems google doesn't think we are gzipping the content: 
![image](https://user-images.githubusercontent.com/2349007/36914701-d27ff0a6-1e13-11e8-8ba8-fcf4e093c32c.png) (from a chrome devtools `audit`)

aappling-usgs:
Yes, David got it :+1: - compression should exist on beta and prod

aappling-usgs:
this issue is also being addressed in #97, #111, probably some others

PR #187, opened 2018-04-16
show county bounds on state load

### LINKED ISSUE #128: Show county bounds when a state is loaded first
If the first page load is to a state view rather than the national view, state boundaries should be visible but aren't:

image

lindsaycarr:
This is an effort to fix #128. I don't think I really understand how the cache stuff in `counties.js` is working because loading just this one state's counties takes almost as much time as loading everything. I have a bunch of `console.log` in the code so that I can watch the order that things happen. Things are happening multiple times though and I think it's because `updateView` calls `updateCounties` which happens at the very beginning when states are added but there is not county data yet so no boundaries show up. Then I make a call to just those state bounds again with my addition of `loadInitialCounties`, and then everything starts loading with `updateCounties('USA')`. I know there's a better way to do this, just not sure what.

Here's how awful it is:

load_county_boundries

2 hrs so far.

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -185,11 +185,9 @@ function loadInitialCounties() {
   // update the view once the county data is loaded
   
   function waitForCounties(error, results){
-    console.log(results);
     updateView(activeView);
   }
   
-  console.log('about to start county load');
   d3.queue()
     .defer(loadCountyBounds, activeView)
     .await(waitForCounties);
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -159,6 +159,10 @@ function fillMap() {
   // add the main, active map features
   addStates(map, stateBoundsUSA);
   
+  if(activeView !== "USA") {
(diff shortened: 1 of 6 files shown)


jread-usgs:
I don't think that is _that_ bad. If the county borders transitioned quickly from transparent to opaque it would look less jarring (same thing w/ the fill). 

Do we know yet how common coming in w/ a state in the URL will be? If, for example, we use it for the url in our state-specific tweets+GIFs, it would end up being somewhat common, but otherwise, maybe a 3% thing?

I'd be in favor of making a small style change (the opacity transition) and leaving as-is. We don't want to go down the path of a complex implementation to solve this if there is still the chance we might make changes to the load ( not sure if there is, but see #62 )

lindsaycarr:
A little better with transitions. Ready to merge.

load_county_boundries

30 more min for total of 2.5 hrs.

Issue #186, opened 2018-04-16
Update rank-em figure functionality


jread-usgs:
once #183 is up and ready, we'll need to:
 - [x] switch the reliance on "hover" to overlap of bounding boxes when a state is dragged over a bar (hover isn't a solution that works for mobile)
 - [x] use d3.scale to set up x and y bar chart placement in the figure
 - [x] let the user know somehow that they were wrong (if they were)
 - [x] [**icing**] link the "closed" bars to their state/territory on mouseover
 - [x] [**icing**] show total value on mouseover of state or bar (when "closed")
 - [x] [**icing**] used cursor ["grab"](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) when on top of the moveable states
 - [ ] moveable states create empty `<use>` elements for locked states

lindsaycarr:
After doing a big text overhaul, I feel like there are two things missing from this figure and would like to think about adding them if possible.

1) Text that says "Drag the states over its matching bar" that goes away as soon as you start dragging.
2) When you successfully complete the exercise, text that explains why Idaho might be a little surprising. E.g. "You might have thought that Idaho would be smaller than Oklahoma and Michigan given it's population, but it has large pockets of agriculture that increase it's total water use." (but better than that wording)

jread-usgs:
great ideas. Similar "drag the .." text type as in climate-fish viz?

lindsaycarr:
Yes, exactly what I was picturing.

jread-usgs:
For font, in the climate fish viz we used 
```xml
<style type="text/css">@import url(https://fonts.googleapis.com/css?family=Shadows+Into+Light);</style
```

lindsaycarr:
Working on some of the recent things brought up after jread's pr #217

lindsaycarr:
Closed by #232

PR #191, opened 2018-04-17
State data

ldecicco-USGS:
image

I think this takes care of #183 and we can now do #186 ...

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -63,6 +63,7 @@ function rankEm() {
         .on("drag", dragging)
         .on("end", dragdone));
     
+  // add states
   
   var barGroups = svgStates.select('#ranked-states-bars')
     .selectAll('g')
@@ -145,7 +146,6 @@ function rankEm() {
 }
   
 
-// add states
 
   
 
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -17,12 +17,20 @@ stateMap.append('g')
 svgStates.append('g')
   .attr('id','ranked-states-bars');
   
-var bardata = [{"wu":80, "abrv":'OK', "open": true}, {"wu":140, "abrv":'WI', "open":false}, {"wu":180, "abrv":'MI', "open":true}, 
-    {"wu":210, "abrv":'ID', "open":true}, {"wu":260, "abrv":'TX', "open":false}, {"wu":310, "abrv":'CA', "open":false}];
+// Read state data and add it to figure
+d3.queue()
+  .defer(d3.json, "data/wu_state_data.json")
+  .await(rankEm);
+  
+function rankEm() {
 
-
-// add states
-  svgStates.select('#ranked-states-moved')
+  // arguments[0] is the error
+	var error = arguments[0];
+	if (error) throw error;
+	
+	var bardata = arguments[1];
+	
+	svgStates.select('#ranked-states-moved')
     .selectAll( 'use' )
     .data(bardata)
     .enter()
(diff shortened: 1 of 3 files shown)


js/rank_states_utils.js
```diff
@@ -17,12 +17,20 @@ stateMap.append('g')
 svgStates.append('g')
   .attr('id','ranked-states-bars');
   
-var bardata = [{"wu":80, "abrv":'OK', "open": true}, {"wu":140, "abrv":'WI', "open":false}, {"wu":180, "abrv":'MI', "open":true}, 
-    {"wu":210, "abrv":'ID', "open":true}, {"wu":260, "abrv":'TX', "open":false}, {"wu":310, "abrv":'CA', "open":false}];
-
+// Read state data and add it to figure
+d3.queue()
+  .defer(d3.json, "data/wu_state_data.json")
+  .await(rankEm);
```
aappling-usgs: This looks functional. But for your d3 edification, another option when you have just one data file is that instead of these 3 lines you can instead do
```js
d3.json('data/wu_state_data.json', rankEm);
```
where then `rankEm` would take a slightly different approach to arguments (they get passed as typical arguments rather than being contained within the `arguments` vector):
```js
rankEm(error, bardata) {
  if(error) throw error;
  svgStates.select....
}
```

viz.yaml
```diff
@@ -176,6 +176,15 @@ process:
       col_names: wu_type_column_names
     processor: summarize_wu_data
     scripts: [scripts/process/summarize_wu_data.R]
+  -
+    id: wu_state_data_json
+    location: cache/wu_state_data.json
+    mimetype: application/json
+    drag_states: ["ID","OK","MI"]
```
aappling-usgs: :+1: for placement of this decision in the viz.yaml. but don't we often put arguments nested within an args holder, e.g.
```
mimetype: applicaton/json
process_args:
  drag_states: ["ID","OK","MI"]
```
?
i don't know if we've actually settled on a convention for this, but i prefer having them nested so i know which info is going to the function and which is getting used by vizlab for other purposes. i realize and have no solution for the fact that things like `location` blur that line.

scripts/process/state_wu_data.R
```diff
@@ -0,0 +1,18 @@
...
+#'
+process.state_wu_data <- function(viz) {
+  
+  deps <- readDepends(viz)
+  wu_df <- deps[["rawdata"]]
+
+  # calculate national sums of each category
+  wu_df_state <- dplyr::select(wu_df, STATE, total) %>%
+    dplyr::group_by(STATE) %>%
+    dplyr::summarise(wu = round(sum(total))) %>%
```
lindsaycarr: couldn't find a Union Jack emoji ..... but how British of you :)

aappling-usgs: :laughing:

ldecicco-USGS: autocomplete...

lindsaycarr:
Yeah, just probably some issues... we will have to set different scaling up for the height of the rectangles in #186. I think we can make a new function like:

```
var scaleBars = d3.scaleLinear()
         .domain(minWU, maxWU)
         .range(0, 250);
```

Then the height of each bar uses that scale.

js/rank_states_utils.js
lindsaycarr: oooo I didn't realize that. Thought we had to nest or queue. Cool!

jread-usgs:
looks good. all set for merge?

ldecicco-USGS:
I think so

PR #192, opened 2018-04-17
Rank em functionality (partial)

jread-usgs:
Looking good !

image

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -1,7 +1,12 @@
 
+var rankSvg = {
+    width: 1000,
+    height: 300
+};
+
 var svgStates = d3.select("#rank-states-interactive")
   .append("svg")
-  .attr('viewBox', '0 0 1000 300')
+  .attr('viewBox', '0 0 '+ rankSvg.width + " " + rankSvg.height)
   .attr('preserveAspectRatio', 'xMidYMid');
 
 var stateMap = svgStates.append('g')
@@ -43,7 +48,6 @@ var bardata = [{"wu":80, "abrv":'OK', "open": true}, {"wu":140, "abrv":'WI', "op
     })
       .classed('draggable','true') 
       .style("fill",categoryToColor("total"))
-      .style("stroke-dasharray",null)
       .attr('id', function(d) {
         return d.abrv + '-rank';
       })
@@ -90,29 +94,12 @@ var bardata = [{"wu":80, "abrv":'OK', "open": true}, {"wu":140, "abrv":'WI', "op
     .classed('closed-rank-bar', function(d){
       return !d.open;
     })
-    .style('fill', function(d){
-      if (!d.open){
-        return categoryToColor("total");
-      }
-    })
-    .on('mouseover', function(d){
-      if (d.open){
-        d3.select(this)
-          .style("stroke-dasharray",null)
-          .classed('chosen-rank-bar',true)
-          .style("fill", categoryToColor("total"));
-      }
-      })
-    .on('mouseout', function(d){
-      if (d.open){
-        d3.select(this)
-          .style("stroke-dasharray","4, 2")
-          .classed('chosen-rank-bar',false);
-        } 
-      });
+    .style('fill', categoryToColor("total"));
   
     function dragging(d) {
-      d3.select(this).attr("transform", "translate(" + (d3.event.x) + "," + (d3.event.y) + ")");
+      var thisShape = d3.select(this);
+      thisShape.attr("transform", "translate(" + (d3.event.x) + "," + (d3.event.y) + ")");
+      overRankBar(thisShape.node().getBoundingClientRect());
     }
 
     function dragdone(d) {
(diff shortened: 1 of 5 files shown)


js/rank_states_utils.js
```diff
@@ -30,6 +35,18 @@ function rankEm() {
 	
 	var bardata = arguments[1];
 	
+	var x = d3.scaleLinear()
+	  .range([0, rankSvg.width])
+	  .domain([0, bardata.length]);
+	
+	var barWidth = rankSvg.width / bardata.length * 0.92;
```
aappling-usgs: We can make even fuller use of D3 for the x scale by using `d3.scaleBand` instead of `d3.scaleLinear`. See https://bl.ocks.org/d3indepth/1aef77d17863e603ff4e84226db5b227 for an example. `scaleBand` allows you to set the padding and total range and get back `bandScale.bandwidth()` (i.e., it computes barWidth for you). And it deals with any x nudging due to padding without you having to set any numbers. The benefits of scaleBand over scaleLinear are pretty minor, but might be worth switching just to practice the D3 way.

js/rank_states_utils.js
```diff
@@ -71,80 +87,84 @@ function rankEm() {
...
+  
+  var openBars = d3.selectAll('.open-rank-bar').nodes();
+  for (var i = 0; i < openBars.length; i++) { 
+    var thisBar = openBars[i];
+    var barBox = thisBar.getBoundingClientRect();
+    //console.log('testing overlap on bar '+bbox.right+' vs '+x);
+    var isOverBar = !(barBox.right < shapeBox.left || 
+                barBox.left > shapeBox.right || 
+                barBox.bottom < shapeBox.top || 
+                barBox.top > shapeBox.bottom);
```
lindsaycarr: is this all to take care of being just barely outside of the rectangle?

js/rank_states_utils.js
```diff
@@ -71,80 +87,84 @@ function rankEm() {
...
-      d3.select(this)
-        .transition().duration(600)
-          .attr('transform','translate(0,0)'); 
+  function overRankBar(shapeBox){
+  
+  var openBars = d3.selectAll('.open-rank-bar').nodes();
+  for (var i = 0; i < openBars.length; i++) { 
+    var thisBar = openBars[i];
+    var barBox = thisBar.getBoundingClientRect();
+    //console.log('testing overlap on bar '+bbox.right+' vs '+x);
```
lindsaycarr: don't forget to remove this console.log()

js/rank_states_utils.js
```diff
@@ -71,80 +87,84 @@ function rankEm() {
...
-
-  function dragdone(d) {
-    var barchoice = d3.select('.chosen-rank-bar');
-    if (barchoice.empty()){
-      d3.select(this)
-        .transition().duration(600)
-          .attr('transform','translate(0,0)'); 
+  function overRankBar(shapeBox){
+  
+  var openBars = d3.selectAll('.open-rank-bar').nodes();
```
aappling-usgs: funky indentation?

lindsaycarr:
I wonder if mobile should change at all ... just keep a few of the states for the bar chart? Those rectangles are going to be pretty small.

image

aappling-usgs:
@lindsaycarr I think keeping all 50 states is important. Also, our national map with callouts is going to take the cake for small details; the bars still look pretty good to me in comparison.

js/rank_states_utils.js
jread-usgs: no, this is to test whether the state's bounding box is overlapping with one of the rectangles that is "open". My last PR had this taking advantage of the bar's hover event, but I changed it to this so it would also work on mobile.

jread-usgs:
Yeah, I'd like to keep all of the states too. If we want to change up the implementation for mobile for this figure, I think we could orient it vertically instead (rotate the plot -90°) to put it in a more "portrait" mode. But let's see how things are on mobile first.

js/rank_states_utils.js
```diff
@@ -30,6 +36,20 @@ function rankEm() {
 	
 	var bardata = arguments[1];
 	
+	var barDomain = [];
+	for (var i = 0; i != bardata.length; ++i) barDomain.push(i);
```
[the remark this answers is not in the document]
aappling-usgs: Hmm, stack overflow seems to say that what you have is about as good as you can hope to get. https://stackoverflow.com/questions/6299500/tersest-way-to-create-an-array-of-integers-from-1-20-in-javascript?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

Gives me the warm fuzzies for R and other vectorization-oriented languages =)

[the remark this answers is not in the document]
jread-usgs: in the end I went w/ https://github.com/USGS-VIZLAB/water-use-15/pull/192/commits/2d0ddee505de2f0d84a12e6e2d77181a09f7ca74 which I think is a better solution for `scaleBand().domain` when we are using the index for x-axis placement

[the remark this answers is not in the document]
aappling-usgs: :+1: - that looks good to me.

Also, I didn't really think about the `domain` part of your question - another option for scaleBand is to pass in categorical values, e.g., STATE_ABBVs, rather than coming up with numeric indices. For example, the xScale in this example https://bl.ocks.org/EmbraceLife/677054c8f535c77ddd95485523d97fcd uses:
```
// set ordinal data's range: array of all data elements
  xScale.domain(data.map(function(d) { return d.name; }));
```

PR #194, opened 2018-04-17
Drop tooltips

### LINKED ISSUE #181: Drop tooltips, add info in category legend
See mockups at https://github.com/USGS-VIZLAB/water-use-15/issues/89#issuecomment-378579693 -- looking to get the state/county/nation name and the water use value into the legend. We can drop population.

Task predicted in #135:
* rounding/styling water use values shown in the category legend (e.g., 2.00000021)

Task worried about by me:
* will long county names overrun the button/legend box? we may need to make the box wider and/or rearrange the text so it fits. See Alaska for some extra-long county names.

ldecicco-USGS:
Fixes #181

You guys will have to settle for screen shots for now:

image

Hover:

image

National view:

image

If you step out of the map, it clears the numbers and title back to "Water Use"

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -51,8 +51,8 @@ function addButtons() {
   // button category use values
   buttons.append('text')
     .classed('category-amount', true)
-    .text(function(d){
-      return 'XX';
+    .attr('id', function(d){
+      return (d + "_button_text");
     })
     .style('font-weight', 'thick')
     .attr('text-anchor', 'end');
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -15,11 +15,11 @@ function addButtons() {
       return d;
     });
     
-  // legend title
+    // legend title
   buttonBox.append('text')
     .classed('title', true)
     .attr('id', 'legend-title')
-    .text('Water Use in XXX');
+    .text("Water use");
   
   // legend subtitle
   buttonBox.append('text')
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -242,18 +242,11 @@ function documentCategorySwitch(category, prevCategory, action) {
 
 var toolTipTimer = null;
 var toolTipDelay = 1000; //ms
+
 function showToolTip(d, category) {
 
-  // change tooltip
-  d3.select(".tooltip")
-    .classed("hidden", false)
-    .style("left", (d3.event.pageX + 35) + "px")
-    .style("top", (d3.event.pageY - 50) + "px");
-  d3.select(".tooltip")
-    .html(d.COUNTY + ", " + d.STATE_ABBV + "<br/>" + 
-            "Population: " + d.countypop + "<br/>" +
-            categoryToName(category) + ": " + 
-              d[[category]] + " " + "MGD");
+  waterUseViz.elements.buttonBox.selectAll("#legend-title").text(d.COUNTY + ", " + d.STATE_ABBV);
+
   if(toolTipTimer){
     clearTimeout(toolTipTimer);
   }
@@ -265,8 +258,7 @@ function showToolTip(d, category) {
 }
 
 function hideToolTip() {
-  d3.select(".tooltip")
-    .classed("hidden", true);
+
   if (toolTipTimer){
       clearTimeout(toolTipTimer); // stop ga for edge states
     }
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -245,7 +245,35 @@ var toolTipDelay = 1000; //ms
 
 function showToolTip(d, category) {
 
-  waterUseViz.elements.buttonBox.selectAll("#legend-title").text(d.COUNTY + ", " + d.STATE_ABBV);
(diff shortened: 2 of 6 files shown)


ldecicco-USGS:
I think my screen capture software is cutting off the cursor....but pretend there's the standard white arrow pointed to those counties

js/buttons.js
```diff
@@ -51,8 +51,8 @@ function addButtons() {
   // button category use values
   buttons.append('text')
     .classed('category-amount', true)
-    .text(function(d){
-      return 'XX';
+    .attr('id', function(d){
+      return (d + "_button_text");
```
aappling-usgs: let's use kebab-case instead of snake_case per https://github.com/USGS-VIZLAB/vizlab/wiki/Style-guide bullet 6

js/map_utils.js
```diff
@@ -242,18 +242,39 @@ function documentCategorySwitch(category, prevCategory, action) {
...
+  waterUseViz.elements
+    .buttonBox
+    .selectAll("#industrial_button_text")
+    .text(d[["industrial"]]);
+
+  waterUseViz.elements
+    .buttonBox
+    .selectAll("#thermoelectric_button_text")
+    .text(d[["thermoelectric"]]);
+
```
aappling-usgs: Getting fancy with the selectors! Do they all need to be edited every time? What if we just update the two text elements that are visible? Also, note that there's a `select()` function at your disposal when you just want to select one thing.
```
waerUseViz.elements.buttonBox
  .select('#legend-title')
  .text(d.COUNTY + ", " + d.STATE_ABBV);

waterUseViz.elements.buttonBox
  .select("#" + category + "-button-text")
  .text(d[[category]]);
```

js/map_utils.js
```diff
@@ -264,9 +285,38 @@ function showToolTip(d, category) {
...
 
-function hideToolTip() {
-  d3.select(".tooltip")
-    .classed("hidden", true);
+function clearLegendText() {
+
+  waterUseViz.elements
+    .buttonBox
+    .selectAll("#legend-title")
+    .text("Water Use");
```
aappling-usgs: I believe this was "Water use" with lowercase u the first time you set it, up in buttons.js line 22. What if rather than setting it at all up there, you simply call `clearLegendText()` from within `addButtons`?

js/map_utils.js
```diff
@@ -264,9 +285,38 @@ function showToolTip(d, category) {
...
+    
+  waterUseViz.elements
+    .buttonBox
+    .selectAll("#thermoelectric_button_text")
+    .text("");
+    
+  waterUseViz.elements
+    .buttonBox
+    .selectAll("#publicsupply_button_text")
+    .text("");
```
aappling-usgs: Here I see that it's useful to set all of the text elements to make sure they're all empty, but there's still a way to replace these 24 lines with 5 lines:
```js
waterUseViz.elements
  .buttonBox
  .selectAll('.category-amount')
  .text("");
```
because `.text` is getting called after a `.selectAll`, it automatically gets looped over all the elements in the selection (in this case all 5 of the text boxes with class `category-amount`)

scripts/process/merge_centroids_wu.R
```diff
@@ -11,8 +12,9 @@ process.merge_centroids_wu <- function(viz) {
     left_join(wu_data_15_simple, by=c('GEOID'='FIPS')) %>%
     select(GEOID, STATE_ABBV, COUNTY, countypop:industrial) %>% 
     rowwise() %>% 
-    mutate(other = total - sum(thermoelectric, publicsupply, irrigation, industrial))
-  
+    mutate(other = total - sum(thermoelectric, publicsupply, irrigation, industrial)) %>%
+    mutate(total = as.integer(round(total)))
```
aappling-usgs: I think we should make a style decision about how many digits to show that allows us to be consistent between total and other categories, and at any spatial resolutions we choose to show. If the raw data come in at 2 digits after the decimal, maybe we always show exactly 2 digits? or we could choose to show all categories as integers; that would be a little cleaner, though I imagine the WU team might be sad about giving up those hard-won final digits.

js/map_utils.js
ldecicco-USGS: Doh! Yeah, I should have thought of that

scripts/process/merge_centroids_wu.R
ldecicco-USGS: Yeah, I was going to address all the rounding in another PR. For whatever reason...it's always trickier (for me?) than it seems. I figured though that integers in total were probably OK (since we know those are the biggest numbers.... and that prevents the numerous 10.0000000000003 )

js/map_utils.js
ldecicco-USGS: Hmm...I had done it like you described, but then when you hovered over the selectors, there were old numbers that were showing up...but THEN I realized no matter what we need the `clearLegendText`..so yeah, maybe we could go back to this scheme...let me try

PR #195, opened 2018-04-18
Rank em

jread-usgs:
rankem_wrong

The last part of MVP on #186 

I assume we'll want to add a little something more in the end (like text), but I think this is what we need for MVP for beta release

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -114,6 +114,7 @@ function rankEm() {
     })
     .attr('width', scaleX.bandwidth())
     .style("stroke-dasharray","4, 2")
+    .style('stroke','rgb(190,190,190)')
     .classed('open-rank-bar', function(d){
       return d.open;
     })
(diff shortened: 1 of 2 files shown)


js/rank_states_utils.js
```diff
@@ -135,17 +136,36 @@ function rankEm() {
...
+            .classed('open-bar-name',false);
+          d3.select(this).style('opacity', 0).attr('transform','scale(0)'); 
+          // one last time in cast it overlapped two rectangles
+          overRankBar(d3.select(this).node().getBoundingClientRect()); 
+          
+        } else {
+          barchoice.attr('class','open-rank-bar');
+          barchoice.transition().duration(600).ease(d3.easeLinear)
+            .style('stroke-dashoffset',"6")
+            .style('stroke', 'rgb(200, 65, 39)') // emphasize with a redorange stroke
```
lindsaycarr: in the gif you included, that color didn't come through obviously. Should we fill the bar to make it more obvious? We could also have it flash redorange fill and then back to nothing.

js/rank_states_utils.js
```diff
@@ -135,17 +136,36 @@ function rankEm() {
...
-        barchoice
-          .on('mouseover', null)
-          .on('mouseout', null)
-          .attr('class','closed-rank-bar');
-        d3.select(barchoice.node().parentNode).select('text')
-          .classed('open-bar-name',false);
-        d3.select(this).style('opacity', 0)
-          .attr('transform','scale(0)'); 
+        
+        var thisBarname = d3.select(barchoice.node().parentNode).select('text').text();
```
lindsaycarr: so that first select becomes the group that each rect and text exist in, right? The `d3.select(barchoice.node().parentNode)`? Just still learning all these ways of selecting things :)

js/rank_states_utils.js
jread-usgs: Yeah, it isn't super obvious, but I didn't want to go too far messing around with strokes and classes since I thought we still might make some changes as to how that is organized. I am not very satisfied with my mix of `.style` and `.classed` in here, but I think it is OK for now.

js/rank_states_utils.js
jread-usgs: Yeah, so each rect is within a "parent" `g` that has the transform in it, and contains rect and text. I use this to select the text element in the same group as the rect that was selected. Not sure if there is a better way to do that. 
here, `parentNode` is the `g` group, and then I select the only text element that belongs to it.

js/rank_states_utils.js
jread-usgs: oh yeah, and it does look like the gif colors made my redorange look darker than it appears on screen

PR #210, opened 2018-04-18
populate legend w national data when view==='USA'

aappling-usgs:
image

@ldecicco-USGS - i wanted to test out what i was suggesting in Slack to make sure I wasn't throwing bad code at you. I think this works! If I haven't messed up anything else, then I think the next task is to link up the state data, which I know you have a start on.

I don't love `updateLegendTextToView` as the new name of `clearLegendText`, but I think it's a slight improvement because now `clearLegendText` is doing so much more than clearing. Feel free to rename to something better.

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -26,8 +26,6 @@ function addButtons() {
     .attr('id', 'legend-subtitle')
     .text('(millions of gallons per day)');
   
-  clearLegendText();
-  
   // button rectangles
   buttons.append('rect')
     .attr('x', 0)
@@ -56,10 +54,12 @@ function addButtons() {
       return (d + "-button-text");
     })
     .style('font-weight', 'thick')
-    .attr('text-anchor', 'end');
+    .attr('text-anchor', 'end')
+    .each( function(d) { this.__data__ = {category: d, wateruse: NaN}; });
   
   // add styling to the buttons and text according to which is active
   updateButtons(activeCategory);
+  clearLegendText();
   
 }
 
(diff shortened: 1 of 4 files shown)


js/buttons.js
```diff
@@ -56,10 +54,12 @@ function addButtons() {
       return (d + "-button-text");
     })
     .style('font-weight', 'thick')
-    .attr('text-anchor', 'end');
+    .attr('text-anchor', 'end')
+    .each( function(d) { this.__data__ = {category: d, wateruse: NaN}; });
```
aappling-usgs: Everything we've appended to `buttons` has automatically inherited the `__data__` attribute of its parent button, which is one element of the `categories` vector, e.g., `'total'`. But for us to be able to merge the `category-amount` text elements with new data when we switch views, we need both the category and the wateruse value to be available, so we need a structure that can match the actual data we'll be passing in later. This line creates that structure by modifying the `__data__` attribute of each `text.category-amount` element.

js/map_utils.js
```diff
@@ -265,17 +265,36 @@ function updateLegendText(d, category) {
...
-    .text("");
+  if(activeView === 'USA') {
+    console.log('activeView is USA, so setting category amounts');
+    
+    waterUseViz.elements
+      .buttonBox
+      .selectAll("#legend-title")
+      .text("U.S. Water Use");
+  
+    d3.json("data/wu_data_15_sum.json",
```
aappling-usgs: we really should load and cache this data somewhere rather than loading it every time. but let's get it working first, then deal with that.

PR #214, opened 2018-04-19
delay for mouseouts

### LINKED ISSUE #160: Revisit need for mouseover delays

lindsaycarr:
This mostly fixes #160 BUT I can't figure out how to rework the `showButtons` function to include the scenario when you mouseout of some category and then immediately back onto the currently selected one. In this scenario, the timeout is cleared out but then on the mouseover the previous and current categories are the same so it skips changing the buttons back.

Here is the before:
category_switching_before

Here is where I'm at:
category_jumps

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -1,3 +1,5 @@
+var mouseTimer;
+
 function addButtons() {
   var buttonBox = waterUseViz.elements.buttonBox = svg.append('g')
     .attr('id', 'button-container');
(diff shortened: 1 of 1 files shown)


lindsaycarr:
I might have a way to do it if I don't clear the timer when the previous mouseover is the same as the current...not sure but can look into it later.

jread-usgs:
I don't have a strong opinion here, but if I had to choose between no mouseover "previews" (onclick only) and the delay, I think I'd choose the onclick. Am I in the minority?

ldecicco-USGS:
I think if it went to only onclick....we'd need to add more text around that really advertised that the user needs to click around. I'd vote we keep it on the table and see if any other strategies come up to allow both.

lindsaycarr:
#215 fixes this issue in a much cleaner way.

PR #215, opened 2018-04-20
for @ldecicco-USGS suggestion on #160

jread-usgs:
For #160. I think this is preferred vs delays, unless we are using those for another reason. It is @ldecicco-USGS 's suggestion of putting a transparent box over the category button to catch the mouse events.

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -26,10 +26,14 @@ function addButtons() {
     .attr('id', 'legend-subtitle')
     .text('(millions of gallons per day)');
   
-  // button rectangles
+  // button rectangles for *style*
   buttons.append('rect')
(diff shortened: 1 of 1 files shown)


lindsaycarr:
I didn't pursue this any more yesterday and went with delays because I was realizing that even with the rectangle behind, you would still be "mousing out" of the button. This works but I don't understand how. That's why I didn't bother with it yesterday.

Also, fyi Alison is not available for reviews until monday (she's heads down on writing today)

lindsaycarr:
Oh wait, I'm digging more and see that it's part of the button group so it's whether you mouseout of the whole group or not. A much cleaner solution than delays.

Issue #203, opened 2018-04-18
Finalize all text


aappling-usgs:
This issue can exclude #161 and #202 but needs to cover all other text in the vizzy.

aappling-usgs:
This issue includes deciding on the text and getting it into the vizzy.

Issue #198, opened 2018-04-18
Improve text layout


aappling-usgs:
Jordan, 4/11: layout needs some work and perhaps some separation between the different concepts. Current view has things kind of floating out there (pie, pie map, placeholder rank-'em states)

Lindsay, 4/18: Reduce header font size and text font size for “how to use the figure” (make it more like a caption). Then wrap text around static pie chart if it looks good

lindsaycarr:
Part of fixing the text up (#203) required me to think about moving things into captions and such. In doing this I tried to add some basic caption-like style and functionality, but that can be totally redone. The point is that you know it's not part of the main body text in order to adjust the layout here :) My PR with my text changes is not yet up, but just wanted to post these thoughts before I forgot.

PR #216, opened 2018-04-20
Fixing text

lindsaycarr:
Fixing #203. I've made some pretty significant changes to the text with this PR. I eliminated quite a bit to minimize how much text we required people to read in between the figures. I'm open to suggestions if people strongly disagree with some of these changes.

Things to note: 
1. I made a small attempt at getting fig caption-like text. No, I don't intend on this being the final look of that text, but I wanted to clearly call it out as "here's something that's not part of the main body text". I expect that #198 will spend more time with the best way to display that (and I already made a comment in there about this very thing).
2. In the opening paragraph, we try to bring context to the 322 billion gallons per day by comparing to the equivalent number of Empire State Buildings. I didn't want to leave an incomplete sentence in there, so I added this although we are currently running it by Nancy to make sure it is good from their perspective.
3. I think that the drag/drop ranking figure could use a little text. I've made that suggestion in #186 because the text would actually need to be incorporated into the figure code and not our mustache files.


layout/templates/key_stories.mustache
```diff
@@ -1,54 +1,62 @@
...
+  <div id="main-story">
+  
+    <h3>How do we use water in the U.S.?</h3>
   
     <div class="key-story-plot" id="national-pie">
-    <p>
-    The U.S. states and territories used X billion gallons per day in 2015. This amount of water is approximately equal to YY (Some percentage of Lake Erie?). We use water in different ways to serve many economically and culturally important needs. The breakdown of different water uses is as follows:
-    </p>
+      <p>
+      The U.S. states and territories used over 322 billion gallons per day in 2015. This amount of water is approximately equal to filling the Empire State Building over 1000 times per day. We use water in different ways to serve many economically and culturally important needs. The national breakdown is as follows:
```
jread-usgs: `Approximately...over 1000 `
I think you can skip approximately, since you are saying "more than"

layout/templates/key_stories.mustache
```diff
@@ -1,54 +1,62 @@
...
+      </p>
+    </div>
+  
+    <div id="fig-caption">
+      <p>
+      Domestic, livestock, aquaculture, and mining compose the "other" category and togther make up 5 percent of total withdrawals.
```
jread-usgs: but here might be a good place to use "approximately"

PR #217, opened 2018-04-23
rank-em nice to haves

jread-usgs:
I didn't get beyond doing a really simple mouseover text implementation. I think we need state long names (or a look up from abbreviations to long names in a global these functions ca access) and the same number formatting we'll use elsewhere. Plus units. We'll want a way to show the text without it being annoying/distracting, which it is now. 

I polished the other parts of this though, including the instructions, the linking between bars/states, and dealing with styles in js instead of mixing parts in css and js. 

ezgif com-video-to-gif 1

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -5,6 +5,9 @@ var rankSvg = {
     bottomMargin: 15,
     updateStyles: function() {
     },
+    liftedState: function() {
+      
+    },
     draggingState: function() {
     },
     droppedState: function() {
@@ -42,6 +45,12 @@ svgStates.append('g')
     .classed('rankem-title', true)
     .attr('text-anchor','middle')
     .text('Drag a state over its matching bar');
+
+svgStates.append('g')
+  .attr('id','rank-data-text')
+  .attr('transform',"translate("+(rankSvg.width * 0.4)+","+(rankSvg.height * 0.42)+")")
+  .append('text')
+    .text(' ');
 
 // Read state data and add it to figure
 d3.queue()
@@ -99,6 +108,7 @@ function rankEm() {
     })
     .datum({x: 0, y: 0})
     .call(d3.drag()
+      .on("start", rankSvg.liftedState)
       .on("drag", rankSvg.draggingState)
       .on("end", rankSvg.droppedState));
     
@@ -153,17 +163,22 @@ rankSvg.updateStyles = function(){
   function clearHighlight(){
     d3.selectAll('.highlight')
         .classed('highlight', false);
+    d3.select("#rank-data-text").select('text')
+        .text(" ");    
       rankSvg.updateStyles();  
+      
   }
   
   lockedBars
     .style('fill', categoryToColor('total'))
-    .style('stroke-width',0)
+    .style('stroke-width', 0)
     .on('mouseover',function(){
       var state = d3.select(this).attr('id').split('-')[0];
       d3.select('#'+state+'-locked').classed('highlight', true);
       d3.select(this).classed('highlight', true);
       rankSvg.updateStyles();
+      d3.select("#rank-data-text").select('text')
+        .text(this.__data__.abrv+": "+this.__data__.wu);
     })
     .on('mouseout',clearHighlight);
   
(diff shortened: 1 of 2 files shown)


js/rank_states_utils.js
```diff
@@ -112,84 +142,170 @@ function rankEm() {
...
+    .style("stroke-dasharray","10, 10")
+    .on('mouseover',function(){
+      d3.selectAll('.highlight')
+        .classed('highlight', false);
+      var state = d3.select(this).attr('id').split('-')[0];
+      d3.select('#'+state+'-bar').classed('highlight', true);
+      d3.select(this).classed('highlight', true);
+      d3.select("#rank-data-text").select('text')
+        .text(this.__data__.abrv+": "+this.__data__.wu);
+      rankSvg.updateStyles();
```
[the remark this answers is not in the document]
jread-usgs: I don't follow. It is part of the chain of things that is executed on mouseover. It should have the same indent level as `d3.select`.., right?

js/rank_states_utils.js
```diff
@@ -112,84 +142,170 @@ function rankEm() {
...
-      }
-      
+rankSvg.updateStyles = function(){
+  var allBars = svgStates.select('#ranked-states-bars').selectAll('rect');
+  var lockedStates = svgStates.select('#ranked-states-locked').selectAll('.locked-state');
+  var draggableStates = svgStates.select('#ranked-states-draggable').selectAll('.draggable-state');
+  var lockedBars = allBars.filter('.locked-rank-bar');
+  var openBars = allBars.filter('*:not(.locked-rank-bar)');
+  
+  function clearHighlight(){
```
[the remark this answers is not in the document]
jread-usgs: I put it in there since it is only used by `updateStyles`. Should it be somewhere else?

js/rank_states_utils.js
```diff
@@ -112,84 +142,170 @@ function rankEm() {
...
+        .text(this.data().wu);
+      rankSvg.updateStyles();
+    })
+    .on('mouseout',clearHighlight);
+    
+  lockedStates
+    .style('fill',"rgb(220,220,220)")
+    .style('stroke',"rgb(190,190,190)")
+    .style('stroke-width', 4)
+    .style("stroke-dasharray","10, 10")
```
aappling-usgs: In other situations I've tried to push any style info I could into css, just switching the class in a call like this one, or into styles.js if required to support animated transitions among styles. My goals have been (1) consolidating styles into 1 or 2 files so it's easy for others to edit them and (2) possibly?? speeding up the code by doing less in d3, more with css. But I don't actually know if 2 is true (could be negligible or maybe even a slowdown), and 1 doesn't feel like a big deal to me especially as more of us become comfortable reading and tweaking d3 code.

So I definitely don't think we need to change this code, but I'm curious about why you chose to put the styling here this time around...maybe if we keep discussing reasoning we'll eventually come up with a unified philosophy on when to put styling where.

js/rank_states_utils.js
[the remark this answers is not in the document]
aappling-usgs: I'm fine with it living here, though I could also see it living as a child of `rankSvg` and sibling to `rankSvg.updateStyles()`.

js/rank_states_utils.js
```diff
@@ -112,84 +142,170 @@ function rankEm() {
...
-            .attr('transform','translate(0,0)');  
-        }
-      }
-      
+rankSvg.updateStyles = function(){
+  var allBars = svgStates.select('#ranked-states-bars').selectAll('rect');
+  var lockedStates = svgStates.select('#ranked-states-locked').selectAll('.locked-state');
+  var draggableStates = svgStates.select('#ranked-states-draggable').selectAll('.draggable-state');
+  var lockedBars = allBars.filter('.locked-rank-bar');
+  var openBars = allBars.filter('*:not(.locked-rank-bar)');
```
aappling-usgs: Cool, I hadn't seen `filter` used with a css selector string before. Looks really handy.

js/rank_states_utils.js
jread-usgs: I ran into some part of this that I had to do in js, so I just slid everything into js so it didn't spill over into both. I think it was the dash-array part that just didn't work well in css-only.

aappling-usgs: :heavy_check_mark: makes sense

Issue #25, opened 2018-02-26
Vizlab-side projection (pre-browser)


aappling-usgs:
The early versions of this viz do projection on the D3 side in the browser. Can we move the projection calculations over to the vizlab side, pre-browser? Could either do it in D3 with node.js, or in R.

aappling-usgs:
http://www.matteoagosti.com/blog/2013/02/24/writing-javascript-modules-for-both-browser-and-node/ looks like a useful reference for using the same custom_projection.js both in spatial processing and in the webpage.

aappling-usgs:
made progress in #224; could run with that and apply preprojection to:
- county and state boundaries
- centroids

jread-usgs:
I don't know that we are taking the whole thing on here (or if this title should be renamed). My understanding is that we are going to attempt to pre-scale AK and PR/VIs before the simplification so that we can: 1) show the VIs on the national view desktop w/o increasing file size too much, 2) further reduce the mobile data footprint, and 3) remove _some_ of the complexity of the browser projection, potentially reducing it to a simple albers one. 

In approaching this issue we may end up needing to compromise on some of the objectives.

jread-usgs:
@lindsaycarr want to weigh in on the scope here?

jread-usgs:
Exploring this now - seems the much easier solution is to avoid the goals of 3) and instead implement a simple scale up/down -> simplify -> scale down/up to original size -> write file. 

This seems to work and we can keep the custom_projection.js in place

lindsaycarr:
Sorry - been in meetings. I think the goal was that we could do what you just said (scale VI/PR up and AK down, then simplify, then scale again) and then treat that resulting topojson as the "medium-sized" json. Then, we could load this medium-res stuff on desktop and mobile as the first thing, and desktop would continue to load high-res in the background.

lindsaycarr:
Hmmm I guess that is not really what this issue is now that I look more closely...I'll get something different going.

lindsaycarr:
See #316

PR #224, opened 2018-04-24
St thomas

### LINKED ISSUE #179: St Thomas county isn't in the zoom
Only St Croix shows up:
180411-st-thomas

aappling-usgs:
Resolves #179, zoom to all 3 Virgin Islands instead of just one.

180423-st-thomas

The problem was that the simplified state boundaries we were using to calculate zoom parameters were too simple - they had simplified away 2 of the 3 Virgin Islands and so were ~correctly zooming to the one VI represented by those bounds.

In hindsight I think I could have simply done the data/callback work I ended up doing in map_utils.js and then computing the zoom scale and x/y values from the high-res state boundaries that would have then been available. But because I didn't see that right away, I ended up also pushing the calculation of centroids and zoom scale into the vizlab preprocessing step; now those parameters are packaged with the hi-res state bounds (where they were calculated). This represents more work than I strictly needed to do to resolve #179 , had I known in advance what was needed, but it's not all bad - I imagine that the preprocessing might make the zoom slightly faster, and also I've learned a lot about how we could pre-project all the coordinates in the entire viz (information we may or may not use for this viz but can definitely use in future vizzies).

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -96,7 +96,7 @@ function zoomToFromState(d, i, j, selection) {
     // if they clicked on a different state, prepare to zoom in
     newView = clickedView;
   }
-
+  
   // zoom to the new view
   updateView(newView);
 }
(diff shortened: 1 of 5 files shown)


scripts/preprocess/custom_projection.js
```diff
@@ -0,0 +1,279 @@
...
+    point: function(x, y) { var i = -1; while (++i < n) streams[i].point(x, y); },
+    sphere: function() { var i = -1; while (++i < n) streams[i].sphere(); },
+    lineStart: function() { var i = -1; while (++i < n) streams[i].lineStart(); },
+    lineEnd: function() { var i = -1; while (++i < n) streams[i].lineEnd(); },
+    polygonStart: function() { var i = -1; while (++i < n) streams[i].polygonStart(); },
+    polygonEnd: function() { var i = -1; while (++i < n) streams[i].polygonEnd(); }
+  };
+}
+
+var custom_projection = module.exports = {
```
aappling-usgs: This second copy of custom_projection.js is necessary almost entirely because of this line (plus line 1, require(d3)). Here we're creating custom_projection as something that node.js will recognize as an exported and therefore `require`able entity, which is why we can call `require('custom_projection.js')` in process_counties_to_states.js (https://github.com/USGS-VIZLAB/water-use-15/pull/224/files#diff-f057c7e7b63321a9e7a46b1a9ca39d1aR23). 

In future vizzies, I think we should figure out how to make custom_projection.js into a module that can also be used directly in a webpage. It appears to be possible - see https://github.com/USGS-VIZLAB/water-use-15/issues/25#issuecomment-383757005 - but seems like a corner we ought to cut for now as our development time for water-use-15 is coming to a close.

scripts/preprocess/process_counties_to_states.js
```diff
@@ -5,26 +5,60 @@
...
+      height: 700
+    }
+  }
+};
+var projection = proj.albersUsaTerritories()
+  .scale([1200])
+  .translate([waterUseViz.dims.map.width / 2, waterUseViz.dims.map.height / 2]);
+  // default is .rotate([96,0]) to center on US (we want this)
+var buildPath = d3.geoPath()
+  .projection(projection);
```
aappling-usgs: Code from line 26 to here is copied from build_map.js; as with custom_projection.js, I think there might be ways to copy less code, but hey, it's Monday night for a viz due Wednesday, so I propose that we settle for having a comment about this copying (line 25).

js/map_utils.js
```diff
@@ -117,43 +117,42 @@ function updateView(newView, fireAnalytics) {
...
+  // record the change for analytics. don't need timeout for view change   
+  if(fireAnalytics) {
+    gtag('event', 'update view', {
+      'event_category': 'figure',
+      'event_label': 'newView=' + newView + '; oldView=' +     oldView + '; category=' + activeCategory
+    });
+  }    
+}
+
+function applyZoomAndStyle(newView) {
```
aappling-usgs: i've pulled the code in `applyZoomAndStyle` out from the body of `updateView` so that it can be nested as a callback to updateStateData (lines 125-127 above, https://github.com/USGS-VIZLAB/water-use-15/pull/224/files#diff-2afd72a00f31d3d53312dce9ff03a875R125). This allows us to ensure that the high-resolution state data, which have the high-resolution=correct zoom parameters, are available for use in zooming.

js/map_utils.js
```diff
@@ -117,43 +117,42 @@ function updateView(newView, fireAnalytics) {
     updateStateSelector(activeView);
   } 
   
+  // update the geospatial data for the upcoming resolution
+  updateCounties(activeView);
+  updateStates(activeView);
```
lindsaycarr: so I still think that a lot of the slowness in zooming comes from the fact that when you zoom in and out repeatedly, are these loading and re-loading data every time? I still haven't figured out how they work

aappling-usgs: They're not re-loading data from the internet every time - the code in `updateStateData` and `loadCountyBounds` caches anything locally that has already been loaded with `d3.json`. 

Also, my local experience is that zooming is no slower than changing categories (another thing I tried and failed to make really snappy), so I'm not sure we should spend too much more energy on this...

Still, if we wanted to work even harder on zoom times, one area of possible (but not confirmed!) inefficiency is that we reproject and reapply the low-resolution county boundaries every time we zoom out. We could try sticking with the high-resolution counties for every state that has them. This would involve doing something different at this line, which initiates the reapplication of the low-res boundaries: https://github.com/USGS-VIZLAB/water-use-15/blob/8597065eae5d3932101a1bdb6a2521e15869be23/js/counties.js#L30. We'd need to tease apart the styling from the path building within `displayCountyBounds` to make this work - I'm willing to chew on this for a few minutes, but not in this PR. And can't guarantee that it will make things faster; might make things slower, so need to check the performance.

Issue #202, opened 2018-04-18
Add the finalized key stories text to the figure


aappling-usgs:
After #161, add the final text to the figure in photoshop

aappling-usgs:
Include desktop and mobile screenshots in the PR for this - we want to know how mobile will look in particular.

aappling-usgs:
Also, because text on mobile especially is likely to be painfully small, the assignee of this issue should take responsibility for finding a way to make it look decent. Jordan has suggested chopping the figure up regionally into multiple figures, which may very well be the only way forward. Might even look nicer on desktop that way, too. Feel free to share multiple prototypes as screenshots/mockups within this issue - they probably will attract several interested discussants :hand: .

lindsaycarr:
Here's a prototype of desktop view. The text is available in a PPT file. Marty - I sent you the powerpoint file and PNG of map in the background on 4/18.

![national_pie_map_key_story_withtext](https://user-images.githubusercontent.com/13220910/38961684-4107510a-432f-11e8-8be1-992659d56a1e.png)

PR #226, opened 2018-04-24
Add the finalized key stories text to the figure #202

mwernimont:
Mobile (my phone)
screenshot_2018-04-24-11-13-09
screenshot_2018-04-24-11-13-15
Desktop
screen shot 2018-04-24 at 11 15 37 am

Talked to @lindsaycarr and we discussed trying to just split the figure in half and stacking the two on mobile.  I think its a huge improvement compared to what it was in the readability department.


lindsaycarr:
Marty came and showed me how it looks on his phone right now and text size feels fine to me. I wish I could pinch and zoom on just that single fig (like zooming into the whole webpage) but we disabled that so that people couldn't do it on the main fig.

aappling-usgs:
Per our group Slack convo, Marty is gonna try to squeeze up the font size on mobile a little more, putting in effort within reason given the approaching deadline.

I really like the overall look of the desktop pie map! Information rich, colorful, and nice placement and shaping of the text blocks.

mwernimont:
screenshot_2018-04-24-13-08-02

mwernimont:
Larger font for the mobile figures and I evened up the split in the country to avoid the size difference shrinking the text on either figure.

PR #228, opened 2018-04-24
IE fix

ldecicco-USGS:
image

Everything's still itty-bitty


lindsaycarr:
marty is working on the itty-bitty part with #198

PR #229, opened 2018-04-24
Improve text layout

mwernimont:
Gave each story a little more space, deemphasized the data and reference section, and created a side by side look for the pie chart and its caption to remove some whitespace on the desktop version.
screen shot 2018-04-24 at 3 39 49 pm
screen shot 2018-04-24 at 2 20 48 pm

### DIFF
--- a/js/national_pie.js
+++ b/js/national_pie.js
@@ -1,10 +1,10 @@
 // code to add a static pie chart of the national view
 
-var width = 1000,
-    height = 400,
+var width = 480,
+    height = 350,
     radius = Math.min(width, height) / 3;
 
-var piearea = d3.select("#national-pie")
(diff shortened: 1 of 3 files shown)


mwernimont:
I'm also kinda just going on make the layout better, so if anyone has more specific ideas or examples of what we want this to look like, I can usually make anything we want to happen.

jread-usgs:
seems you are headed in the right direction. Keep on keeping on.

aappling-usgs:
Are the Data Collection and References section responsive - e.g., do they stack up vertically on mobile?

lindsaycarr:
Yep!

image

aappling-usgs:
Resolves #198

PR #232, opened 2018-04-24
Final rank 'em edits

lindsayplatt:
I believe this checks all of the last boxes for #186.

- nicer text for displaying state and value
- change pointer to indicate draggable states (doesn't "grab" the state, but shows that it's interactive)
- state long names added
- units shown
- I _think_ it's taking advantage of Laura's nicer number formatting which happens in R before we even get it. They appear to match the numbers for state totals in the main fig.

rank_em_edits

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -161,6 +161,11 @@ function rankEm(barData) {
         if (svgStates.select('#ranked-states-bars').selectAll('rect').filter('*:not(.locked-rank-bar)').empty()){
           d3.select('#rank-directions')
             .select('text').remove();
+          d3.select('#rank-explanation').selectAll('text')
+            .transition()
+            .delay(600)
+            .duration(600)
+            .style('opacity',1);
         }
   
       } else {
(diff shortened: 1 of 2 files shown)


aappling-usgs:
Looks nice! I think I'd prefer it without the colon after the state name - could you be persuaded?

lindsaycarr:
Ha yeah, that's fine with me!

mwernimont:
That looks great!

jread-usgs:
Looks great. 
The use of `!rankSvg.isDragged` shuts off the highlighting of bars and states while the draggable state is being dragged. I think it would be a good idea to also shut off the legend text during this time.

jread-usgs:
Not sure I have a strong opinion on this, but the previous version showed the total of the three "open" bars when you moused them, but w/o the state name. Do you think it is better to remove that info for the open bar?

aappling-usgs:
Maybe even use "???" in place of the state name when you mouse over the open bars?

js/rank_states_utils.js
```diff
@@ -209,11 +218,29 @@ function rankEm(barData) {
...
-  svgStates.append('g')
+  var labelTextGroup = svgStates.append('g')
     .attr('id','rank-data-text')
-    .attr('transform',"translate("+(rankSvg.width * 0.4)+","+(rankSvg.height * 0.42)+")")
-    .append('text')
-      .text(' ');
+    .attr('text-anchor','middle')
+    .attr('display', 'none');
+  
+  labelTextGroup.append("text")
```
jread-usgs: if all of these are being bumped the same way (x = width.0.6 etc), you can put it in as the transform=translate(width0.6,height0.3) and remove the x and y attributes from the child elements that follow (default in svg is x and y = 0).

lindsaycarr:
OK, I've gone ahead and implemented the shared position for text via translate and the `???` for state name when you hover over open bars. I'm not sure what to do with the `!rankSvg.isDragged` suggestion, because that doesn't appear to be the current behavior. I kind of like being able to hover and see the legend text as I am dragging.

lindsaycarr:
Ended up changing so that legend went away while you dragged, and forgot that originally I had wanted a message to pop up at the end to complete the story. Here's that added:

rankem_explanation

aappling-usgs:
:+1: :+1:

PR #233, opened 2018-04-24
vizlab version update

mwernimont:


viz.yaml
```diff
@@ -14,7 +14,7 @@ info:
   required-packages:
     vizlab:
       repo: github
-      version: 0.3.2
+      version: 0.3.4
       name: USGS-VIZLAB/vizlab
       ref: 6bcdf08319d00f73b0b54ecfb73d1de00837a2de
```
aappling-usgs: Could you update the ref, too? It's c0889d88e9df47dbb763be2bf335c6de8d7b57fd now (latest commit from vizlab now that i've merged your latest PR)

mwernimont:
Okay, I see where you got that guy from.

aappling-usgs:
Cool. I also just created a vizlab release, `v0.3.4`, which we could use as the ref instead, but i think the commit hash works just as well, so let's keep that.

Issue #211, opened 2018-04-18
Mobile legend placement is bad on fresh load


aappling-usgs:
It's getting placed below the map but too far to the left (more like the x value I'd expect if it thought we were in desktop mode)

lindsaycarr:
Do you think this will be helped by improvements made with #190?

aappling-usgs:
resolved by #236

PR #236, opened 2018-04-25
Mobile: Only load hi-res counties, get buttons centered

### LINKED ISSUE #188: Only load hi-res county boundaries in mobile
As of #184, we shouldn't need to load lo-res county boundaries at all in mobile, because at the national view they're neither visible nor interactive. We only need the hi-res boundaries, so users can see and tap on counties once they're zoomed in.

aappling-usgs:
Fixes #188, #211

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -179,7 +179,14 @@ function fillMap() {
   
   // load county data, add and update county polygons.
   // it's OK if it's not done right away; it should be loaded by the time anyone tries to hover!
-  updateCounties('USA');
+  // and it doesn't need to be done at all for mobile
+  if(waterUseViz.mode !== 'mobile') {
+    updateCounties('USA');
+  } else {
+    // set countyBoundsUSA to something small for which !countyBoundsUSA is false so that 
+    // if and when the user zooms out from a state, updateCounties won't try to load the low-res data
+    countyBoundsUSA = true;
+  }
   
   // Read national data and add it to figure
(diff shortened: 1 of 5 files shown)


js/build_map.js
```diff
@@ -179,24 +179,37 @@ function fillMap() {
...
   // it's OK if it's not done right away; it should be loaded by the time anyone tries to hover!
-  updateCounties('USA');
+  // and it doesn't need to be done at all for mobile
+  if(waterUseViz.mode !== 'mobile') {
+    updateCounties('USA');
+  } else {
+    // set countyBoundsUSA to something small for which !countyBoundsUSA is false so that 
+    // if and when the user zooms out from a state, updateCounties won't try to load the low-res data
+    countyBoundsUSA = true;
+  }
```
lindsaycarr: I'm really confused based on the title of this, and the commit title `only load lo-res on desktop`. 

So `updateCounties('USA')` is the thing that loads hi-res data, right? So here we have decided that hi-res is only loaded on desktop and not on mobile. The PR title and one of the commit title's seem to contradict what the actual code is doing. Also confused about what you mean about setting `countyBoundsUSA` to "something small". How will "something small" make `!countyBoundsUSA === false`?

aappling-usgs: `updateCounties('USA')` loads the low-res data. `updateCounties([statename])` loads high-res data.

`true` is "something small" in that it doesn't take much space in memory. and `!true === false`.

lindsaycarr: OHHHH

Issue #234, opened 2018-04-24
Mobile version of Rank'em?


jread-usgs:
Not sure if this is needed, but a quick preview on mobile makes me think the orientation of our fig and scaling will be hard to see/read in mobile portrait mode.

By having a mobile version of the `scaleX` and `scaleY` and text placement should allow us to shift this fig around and keep all of the other interactions the same. 

Maybe worth it?

![image](https://user-images.githubusercontent.com/2349007/39216819-d11198de-47e2-11e8-9f2b-2299ed2deb52.png)

aappling-usgs:
That does look pretty slick in that arrangement....I'm not opposed, depending on time/people availability.

Issue #201, opened 2018-04-18
Test & fix drag/drop figure on mobile


aappling-usgs:
Follows #186 and a push to beta

aappling-usgs:
Looking on beta right now - on my Android Moto G5S Plus, when I attempt to drag a state I end up selecting text (usually the state bar labels in the rankEm figure) and can't get the state to drag. The map is also very small for selecting a state, which would probably be helped by #234 .

On the bright side, tapping on states or bars in the figure works beautifully.

jread-usgs:
I was able to eventually grab a state, but I otherwise had the same problem, hence #234 
I think we should treat the text the same as other svg text and add css properties to make it so you can't highlight it. Should be an easy fix for that.

lindsaycarr:
Just tested this with most recent changes on beta. So much better! Comfortable with declaring this complete!

PR #238, opened 2018-04-25
Mobile rank

jread-usgs:
for #234 

I have some of lindsay's commits in here and probably need to resolve conflicts. 

I also don't know how to tap into the "isMobile", so I created a variable to use in the meantime (we'd replace `rankSvg.mobile` w/ a function call or something else. 
image

I don't think what we have right now for this fig is usable on mobile.


js/rank_states_utils.js
```diff
@@ -179,18 +191,29 @@ function rankEm(barData) {
     }
     rankSvg.updateStyles();
   };
-  
+  var width = rankSvg.desktopWidth;
+  var height = rankSvg.desktopHeight;
+  if (rankSvg.mobile){
```
lindsaycarr: I think you can grab `waterUseViz.mode` which will be either `desktop` or `mobile`. That's a global var, so it should be available here, too.

lindsaycarr:
Wow, Jordan! Jumping ahead and already implementing a solution to #201! Yep, I was going to work on this very issue today. This looks good to me. A little confused on the order in which things should get merged since this has some of my PR in it.

aappling-usgs:
I just merged Lindsay's PR and see that there are no merge conflicts here, which is a relief. I guess the other possibility is that something from Lindsay's PR might be unintentionally overwritten by this PR? Jordan's branch for this PR is 6 commits ahead, 7 commits behind USGS-VIZLAB:master...my guess is that it's all good, but Lindsay, maybe you could read over the changes one more time to make sure nothing happened?

Given that Jordan is at the TRAC meeting today, I'd be happy to create commits and a second PR that takes care of using `waterUseViz.mode` instead of `rankSvg.mobile` after we get this PR in.

aappling-usgs:
@jread-usgs , does "I don't think what we have right now for this fig is usable on mobile." mean you think this is _less_ usable than the desktop version we had yesterday, or just that it's no more usable than that one was? If the former, what is different that needs to be addressed?

lindsaycarr:
Agree, I'm fine with moving ahead on this without Jordan implementing requested changes and having one of us do them. I am slightly concerned that it is going to write over the commits I just added to these files that aren't shown in this PR. Would it be easiest to close this and just copy the small amount of code that actually made this change?

jread-usgs:
for this `I don't think what we have right now for this fig is usable on mobile.` I meant what we have up on beta. Sorry, was unclear.

lindsaycarr:
Ok and with that, Jordan has merged from upstream and it looks all fine. Let's merge this and then deal with `waterUseViz.mode`.

PR #239, opened 2018-04-25
adding a space

mwernimont:
I noticed a part of the text in the mobile pie chart map was missing a space and looked not very professionally proofed.  So I fixed my mistake.


lindsaycarr:
Cool, good catch! One thing I did notice about these is that the resolution still seems kind of low. Any hope for changing that?

lindsaycarr:
Merging this in the meantime, we can think about resolution later.

PR #240, opened 2018-04-25
Variables for resize.js

mwernimont:
@aappling-usgs created the new variables, I wasn't sure if we were bringing in Modernizr for sure and you just asked me to create the variables.  Sorry, I'm so lost on this task, just not clicking with it as much as you and Lindsay seem to be.


mwernimont:
If we do want modernizer in, I can make that happen and set up some hasEvent variables for functions we want to have separated.  Also has a slick mq (media Query functionality as well)

aappling-usgs:
@mwernimont - great, `interactionMode` is a better variable name than what I suggested :+1:. Yes, let's get Modernizr in there next - awesome that you see a way to get that going.

mwernimont:
Searched the repo for waterUseViz.mode and changed them accordingly.  On a few of them I kinda guessed viewport or interactionMode, so let me know if you disagree with any of them.

aappling-usgs:
-

aappling-usgs:
oops sorry wrong button

js/styles.js
```diff
@@ -94,7 +94,7 @@ function resetState() {
     .duration(750)
     .style("stroke-width", 1); // use null to get back to CSS
   
-  if(waterUseViz.mode === "mobile") {
+  if(waterUseViz.viewport === "narrow") {
```
aappling-usgs: i think this one should be `if(waterUseViz.interactionMode === "tap")`

js/build_map.js
```diff
@@ -180,7 +180,7 @@ function fillMap() {
   // load county data, add and update county polygons.
   // it's OK if it's not done right away; it should be loaded by the time anyone tries to hover!
   // and it doesn't need to be done at all for mobile
-  if(waterUseViz.mode !== 'mobile') {
+  if(waterUseViz.viewport !== 'narrow') {
```
aappling-usgs: this one should be `if(waterUseViz.interactionMode === "tap")`

mwernimont: I'm confused on this one cause you are making sure it's not mobile, so !== 'tap'?

viz.yaml
```diff
@@ -307,6 +308,10 @@ publish:
     id: dropdown_selector
     location: "js/dropdown_selector.js"
     mimetype: application/javascript
+  -
+    id: modernizr
+    location: "js/modernizr_custom.js"
+    mimetype: application/javascript
```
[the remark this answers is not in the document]
mwernimont: This is only a subset of the library, as I only picked the features I thought would work for this particular task.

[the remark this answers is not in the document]
mwernimont: Just for future reference

PR #244, opened 2018-04-25
disable text select within drag/drop fig

lindsaycarr:
@aappling-usgs noticed that when you tapped and held your finger down for more than one second and weren't right on a draggable state before starting to drag, it began to highlight the text rather than drag state around. This was very frustrating, so this PR turned off the ability to highlight (as in highlight to copy and paste) text within the rank 'em figure itself.

### DIFF
--- a/layout/css/map.css
+++ b/layout/css/map.css
@@ -3,6 +3,18 @@
   fill: transparent;
 }
 
+#rank-states-interactive text{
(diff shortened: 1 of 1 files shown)


mwernimont:
Look at all dem prefixes.  CSS can be so gross looking sometimes.  But looks great to me!

PR #245, opened 2018-04-25
add sessionID and timestamp to analytics

wdwatkins:
I haven't implemented this for all events yet, so don't merge.  I wanted to get a review first.

I've added the `getTimestamp` and `getSessionId` functions.  `getTimestamp` is a hit-level dimension, so will record the time of every event.  `getSessionId` just needs to be a unique random number that represents the session.  As implemented now, it will be different for each hit, but because it is scoped as a session dimension, Google will only store the last value it receives and apply that to the entire session.  I wanted originally to just make `sessionId` a fixed variable, since it doesn't need to change, but I was getting errors with it being undefined.  This is how it was done here: https://www.simoahava.com/analytics/improve-data-collection-with-four-custom-dimensions/#3-session-id so may be unavoidable.


js/map_utils.js
```diff
@@ -98,10 +97,13 @@ function zoomToFromState(d, i, j, selection) {
   }
   
   // zoom to the new view
-  updateView(newView);
+  updateView(newView, fireAnalytics = true);
```
wdwatkins: analytics weren't firing at all on `updateView()`

js/map_utils.js
```diff
@@ -98,10 +97,13 @@ function zoomToFromState(d, i, j, selection) {
...
 }
 
-function updateView(newView, fireAnalytics) {
+function getSessionId(){return new Date().getTime() + '.' + Math.random().toString(36).substring(5)}
```
aappling-usgs: session ID is a timestamp?

wdwatkins: With some appended random characters.  It's just a way of ensuring a random number.

wdwatkins: To be more specific, it is a timestamp in milliseconds since 1970-01-01

aappling-usgs: Cool, thanks :+1:

PR #248, opened 2018-04-25
adjust height of button box

### LINKED ISSUE #243: On mobile, box for legend/categories needs to expand to cover title

lindsaycarr:
Fix #243. Before image is in the issue. Here's after:

image

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -70,7 +70,7 @@ function addButtons() {
 function resizeButtons() {
   
    // reserve the top band for titles
-   titlesHeight = waterUseViz.dims.buttonBox.height * 0.2;
+   titlesHeight = waterUseViz.dims.buttonBox.height * 0.22;
   
   // recompute the button heights and positions for the new buttonBox width
   var buttonY = d3.scaleBand()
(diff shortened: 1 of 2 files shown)


js/buttons.js
```diff
@@ -111,13 +111,9 @@ function resizeButtons() {
   waterUseViz.elements.buttonBox.selectAll('text.title')
     .attr('x', +activeButton.attr('x'));
   waterUseViz.elements.buttonBox.select('#legend-title')
-    .attr('y', function(d) {
-      return titlesHeight * 0.3;
-    });
+    .attr('dy', '1.2em');
```
aappling-usgs: i don't understand - how can we delete the `attr('y'` calls and have this still work?

aappling-usgs: Oh, wait, i think i get it - `y` is 0 by default. Does `dy` enable the use of `em` units whereas `y` wouldn't let you do that? And does this method do any better on IE than the previous one?

lindsaycarr: yes to the 0 thing, and I've just seen things that need to be on one line above the other use the `dy` and `em` unit convention before. Not sure about IE. Let me check

lindsaycarr: IE
image

aappling-usgs: Doesn't look worse than before; better if anything. I wonder if the vertical centering of the button labels is due to IE having a different default for how text gets anchored vertically? Since you have a preview running, would you be up for trying adding `.style('text-anchor', 'middle')` and/or `.style('dominant-baseline', 'central')` to this block in buttons.js (~line 125)?
```
  waterUseViz.elements.buttonBox.selectAll('.button .category-label')
    .attr('x', +activeButton.attr('x') + activeButton.attr('width') * 0.05)
    .attr('y', function(d) {
      return buttonY.bandwidth()/2;
    });
``` 
internet says that might help: https://stackoverflow.com/questions/5510432/how-to-center-svg-text-vertically-in-ie9?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

lindsaycarr: I've tried `'text-anchor', 'middle'`, `'vertical-align', 'middle'`, `'text-align', 'middle'`, and `'dominant-baseline', 'central'`. None of these had any effect. Only way I could shift anything was by adding `dy` but that also impacted chrome negatively...

aappling-usgs: Hmm, well, I guess leave it for the style sweep (or forever)

image

PR #249, opened 2018-04-25
IE doesn't like default arguments

aappling-usgs:
Someday I'll learn to recognize this problem while reviewing PRs or creating my own code. Not today, apparently. Anyway, this PR gets IE working again.


aappling-usgs:
merging myself because it's wednesday-due-day and this is a short PR and you two have gone home.

PR #251, opened 2018-04-25
better w-o-w comparison

### LINKED ISSUE #203: Finalize all text

lindsaycarr:
Fixes #203 once and for all!


aappling-usgs:
Hah, text is never final :imp:.

PR #256, opened 2018-04-26
Quick fix to get embed to work

ldecicco-USGS:


ldecicco-USGS:
The only error in embed now is:
```
gtag('config', UA-78530187-12);
```
UA is not defined.

PR #267, opened 2018-04-27
fixes #265 - rank-em style feedback

### LINKED ISSUE #265: Finesse rank-em interactivity
- [x] when complete w/ all three states, turn dashed lines to solid, maybe transition lines to blue too
- [x] on mobile, make sure when you start the drag event, that the other highlighted states clear out (text included). Currently if you tap one locked state and start dragging another, the locked state stays highlighted and the text stays while you drag the other
- [x] remove mouse events on the dashed stroke for the empty box. If you mouse over that line you can get the bar into a funky blinking state since the "moused" style has a zero width stroke and the unmoused is > 0.
- [x] NWULT points out that MI and OK seem random as choices. Do we have stories for any other states that we could add (and swap out that state for OK) so we have one high-use story, one low-use story? AZ is higher & more agricultural than you might expect for a desert...

jordansread:


js/rank_states_utils.js
```diff
@@ -83,6 +84,7 @@ function rankEm(barData) {
   openBars
     .style('fill', "rgba(255, 255, 255, 0.0)")
     .style('stroke',"rgb(198,198,198)")
+    .style('pointer-events', 'fill')
```
lindsaycarr: It is still doing the blinking thing, but this is minor, so I don't think we should spend much more time fixing it.

lindsaycarr: Not blinking -- it's because of the bug that I introduced with being able to cheat and hover over the open bar names and see the answer.

Issue #263, opened 2018-04-27
Unify number rounding and commas


aappling-usgs:
Use consistent rounding and commas in numbers in figures 1 and 4.
- Some options / conflicting suggestions from NWULT and JREAD, so choose a rounding strategy.
  - NWULT: Would prefer that these State totals were rounded to 3 significant figures if possible, or maybe just to integers if sig fig is too much. (This would also be good on the US totals at the top.)
  - NWULT: If can’t round, need to display to uniform number of decimal places--note that NY has no decimals, others have 1 or 2. It looks like Excel behavior where it drops all trailing zeros...
  - NWULT: It would make it easier for users to parse the larger values if the display can use commas in the State total values, as is done on the national total values at the top figure.
  - JREAD: add comma to big numbers and round big numbers to whole numbers (don't round DC, for example)

- [ ] County (first) map legend: 
- [ ] State ("rank-em") map legend: round as in fig 1 and use commas as in fig 1

jread-usgs:
My comment started out as a suggestion to display the state totals as integers, but then I realized DC would be 0, which is kind of misleading. I don't agree that trailing zeros need to be kept for this kind of display, but I don't feel too strongly about that. I would suggest we are consistent with the state zoom (no county selected) on fig 1, which itself might need some adjustment (note CA doesn't have a comma, but the national total does).

PR #269, opened 2018-04-27
Cleaning up numbers

ldecicco-USGS:
States now have commas:

image

image

Any state less than 2 gets some sigfigs:
image

There are now 3 significant figures for the county data, but no commas on that.


ldecicco-USGS:
Fixes #263

Issue #271, opened 2018-04-27
Do something (or not) about blips when circles go to zero.


lindsaycarr:
This happens because as the radii got to zero during the transitions and has something to do with the strokes. Since we are using one path for all circles (which does feel A LOT snappier but I didn't want to spend time on actually measuring performance), we cannot figure out which ones will go to zero and make their stroke transparent to prevent that flash. 

So, we have two options 1) keep the strokes and deal with the flash, or 2) remove strokes altogether and not have that weird blip thing happen. I've attached gifs depicting this. To be honest, not having stroke outlines on the circle is not quite as crucial now that we have circles scaling on zoom - there's not quite as much overlap. The other thing to consider is that without strokes, the national map looks very bare because the tiny tiny ones are even less visible.

@jread-usgs and @aappling-usgs what do you think?

jread-usgs:
Agreed. Strokes are not worth it

jread-usgs:
fixed?

lindsaycarr:
Yep, we removed strokes in #289 which was the issue.

PR #272, opened 2018-04-27
Circles as paths

### LINKED ISSUE #119: Use multi-circle svg path for each water use category
Generate paths something like https://github.com/USGS-VIZLAB/water-use-15/pull/118/files#r178587103, but we just need one final path per category

At least for the time being (unless and until we do #25), do this in javascript using the existing centroid data and `projection()`. There are handy functions like `Array.prototype.join()` that should help.

Then draw circles using these paths rather than by drawing each circle separately.

lindsaycarr:
Fixes #119. This all worked perfectly fine for me. I just uncommented the code that Alison had written before and then added the tricks for rescaling the radii on zoom. I tried to limit the number of times the paths were being recalculated, so I never overwrote the national view radii. I only recalculated the paths for the current category when you zoom, or if you are not in the national view when you change categories. I don't think it takes much to recalculate, but just wanted to limit it as much as possible. 

I did look into the weird flashing that David mentioned on Slack (and Laura has mentioned before). Comments captured here #271 (I don't want this issue to be held up by that discussion).

I was not able to locally test this on mobile. Need a mac to do that.

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -180,10 +180,10 @@ function fillMap() {
   
   // add the circles
   // CIRCLES-AS-CIRCLES
-  addCircles(countyCentroids);
+  //addCircles(countyCentroids);
   // CIRCLES-AS-PATHS
-  /*var circlesPaths = prepareCirclePaths(categories, countyCentroids);
-  addCircles(circlesPaths);*/
+  var circlesPaths = prepareCirclePaths(categories, countyCentroids);
+  addCircles(circlesPaths);
   updateCircleCategory(activeCategory);
   
   // manipulate dropdowns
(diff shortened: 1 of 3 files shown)


aappling-usgs:
The code looks solid to me, and yes, it's snappier and I wish we could use it...it's just that I'm still seeing blank maps here and there, most often for Total.

I guess I'll go with tentative approval & merge, knowing that we know how to revert this, so that we can test on more devices on dev. But I'm going to keep #119 open because we need to test and decide whether we can keep this.

lindsaycarr:
Man, that's so bizarre. I hope that's the only issue, but then I also want to know why in the world that's the case!

lindsaycarr:
So this still does that to you on dev?

Issue #259, opened 2018-04-26
Unify css/d3 resize transitions


aappling-usgs:
unify the thresholds for font changes (css) and button locations (d3). The current numbers for this are `@media screen and (min-width:600px)` in custom.css and `if(Modernizr.mq('(min-width: 425px)'))` in resize.js.

Symptom: At the first mobile transition (screen width 750 px), the text in the category legend is too large to fit (up until the 530px transition). Looks particularly bad when thermo category is selected

aappling-usgs:
Yes, those two numbers should be the same. We should see dropdowns and buttons on the bottom, or no dropdowns and buttons on the left.

Testing on mobile would be good - ask Marty or Jordan to set you up or to test it for you? Or if your changes are easily reversible, I'm OK with pushing to dev & beta today as long as you then test on mobile and submit a 2nd PR before the end of the day if needed.

wdwatkins:
And I'm not seeing the symptom you're noting there using devtools.

wdwatkins:
Ah ok, I wasn't clear on the dropdowns/buttons thing.

aappling-usgs:
Next step: test on beta. Beta is building right now.

lindsaycarr:
Looks good on beta to me.
![good_resizing](https://user-images.githubusercontent.com/13220910/39387295-6c8a81f6-4a3e-11e8-9fe1-e440740b8320.gif)

PR #275, opened 2018-04-27
transition buttons and dropdown at same width

wdwatkins:
For #259


mwernimont:
Looks good to me, and makes sense!

Issue #257, opened 2018-04-26
Legend adjustments


aappling-usgs:
- [x] Can the "off" category bar widths be only slightly shorter than the active category bar? Instead of much shorter (see Marty’s mock-up)
- [x] ~have category transitions include a bar width shrink/expand w/ the same duration as the dot transition - this is a d3 thing but is as simple as finding the command where the bar lengths get changed and then inserting `.transition.duration(500)` (or whatever the right duration value is) between the selection and the style/attr calls in that command.~ moved to #300 
- [x] in mobile portrait mode, the left (category name) and right (category value) buffers in the rectangle are too large. they should be smaller and the rectangles should be ~20% taller. also move buttons up into the map a little to make them fit (reduce the buffer between map and buttons, and/or reduce the map height + expand the zoom)
![image](https://user-images.githubusercontent.com/12039957/39336460-810f18b2-496c-11e8-9d3e-f94bb6340cb9.png)
- [x] try having the "off" categories show up as colored rect borders instead of grey fill? We are grey-heavy right now, and grey also gives the impression that you can't toggle something on. Try with transparent and grey fill, see what looks good
- [x] increase opacity on the category legend background rectangle

ldecicco-USGS:
I don't have @mwernimont 's mockup....so not 100% sure what the first checkbox is asking.

I've done this so far:
![image](https://user-images.githubusercontent.com/1105215/39371358-15fadc32-4a07-11e8-82eb-0ee2c0a34501.png)

So far so good?

mwernimont:
That's it exactly.  I just didn't like width difference.  Much better!

aappling-usgs:
Ooh, that works visually! The white buttons are still clickable as expected, right?

ldecicco-USGS:
Yup, it's just the "filled-button" that the style is adjusted, the user events all happen on the invisible "mouser-button".

ldecicco-USGS:
I can't believe how hard I'm struggling to get the opacity to update. I think I'm only updating the opacity of the stroke maybe? It looks like this:

![image](https://user-images.githubusercontent.com/1105215/39373537-868cf632-4a0d-11e8-8b82-6ab6b7b51f65.png)

Here's one of the million iterations I've tried:
```
  waterUseViz.elements.buttonBox
    .selectAll(".button rect")
    .filter(".filled-button")
    .style('opacity', 1);
```
When I inspect, it says:
[...]

PR #276, opened 2018-04-27
Legend style

ldecicco-USGS:
image

I think this covers the 1st, 4th and 5th checkbox in #257.

I'm a noob at both mobile issues (3rd checkbox) and transitions (2nd checkbox)...so I expect those will take me longer.


ldecicco-USGS:
legend

So actually realized what a transition was and, cool, that's neat

ldecicco-USGS:
Well shoot, now that I made that fancy gif, I just reloaded and it's doing the initial transition backwards...hold on reviewing this....

ldecicco-USGS:
OK, still doesn't work...I managed to lose the state zoom.

js/styles.js
```diff
@@ -9,6 +9,16 @@ function categoryToName(category) {
   else { return "none"; }
 }
 
+function getColor(category){
```
jread-usgs: I think the categoryToColor function should take an optional second argument, which is the alpha value you want. Then we can use a single function.

aappling-usgs: Good idea - and @ldecicco-USGS remember to use your IE-compatible pattern for optional arguments! (I'm learning from my mistake yesterday =) )

js/styles.js
```diff
@@ -9,24 +9,26 @@ function categoryToName(category) {
...
+   if (category === "total") { return "38, 140, 178"; }
+    else if (category === "thermoelectric") { return "237, 201, 72"; }
+    else if (category === "publicsupply") { return "118, 183, 178"; }
+    else if (category === "irrigation") { return "89, 161, 79"; }
+    else if (category === "industrial") { return "225, 87, 89"; }
+    else if (category === "other") { return "169, 169, 169"; }
+    else { return "none"; }
+}
+
+function categoryToColor(category, alpha) {
```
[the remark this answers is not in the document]
ldecicco-USGS: That seems harder for me to fiddle with that not including it...

js/styles.js
```diff
@@ -9,24 +9,26 @@ function categoryToName(category) {
   else { return "none"; }
 }
 
-function categoryToColor(category, stroke) {
-  
-  if(stroke === undefined) {
-      stroke = false;
-   }  
+function getColor(category){
```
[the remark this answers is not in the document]
jread-usgs: Almost all of the styles are handled in js -- and those that aren't in js are in CSS I think -- so I don't think it helps us much to remove this one and put it in the yaml.

js/buttons.js
```diff
@@ -132,25 +153,39 @@ function resizeButtons() {
...
-        return '#c6c6c6';
+        return 'transparent';
       }
     });
-  updateButtonWidths(category);
+    
+  waterUseViz.elements.buttonBox
+    .selectAll(".button rect")
+    .filter(".filled-button")
+    .style('opacity', function(d){
```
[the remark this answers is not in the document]
ldecicco-USGS: Yup, this was a pre-alpha argument fix

viz.yaml
```diff
@@ -463,6 +464,7 @@ publish:
       county_boundaries_mobile: county_boundaries_mobile_topojson
       county_centroids_wu: county_centroids_wu_tsv
       wu_data_15_range:  wu_data_15_range_json
+      wu_state_data_json: wu_state_data_json
```
ldecicco-USGS: So I have literally no idea how dev/beta are working without this line.

jread-usgs: if this file is necessary (looks like it is!) dev/beta may be working because the file was built and put in `target` on an earlier version and then it never ends up getting deleted.

js/buttons.js
```diff
@@ -153,14 +153,15 @@ function updateButtonWidths(category) {
...
   waterUseViz.elements.buttonBox
-    .selectAll('.button rect')
+    .selectAll('.button .filled-button')
     .style('fill', function(d) {
       var col = categoryToColor(d,0);
       if(d === category) {
         col = categoryToColor(d,1);
       }
-      return (col);
+      return(col);
```
jread-usgs: does this work? I thought javascript was `return value;` instead of R's `return(value)`

ldecicco-USGS: I just double checked and it seems to work both ways just fine. I'll take out the parenthesis for style consistencies...but yeah, it works.

jread-usgs:
On last thing for style - can you change the stroke to 1?

ldecicco-USGS:
My old lady eyes prefer fatter lines, but sure:

image

jread-usgs:
and I think the fill alpha of the buttons needs to be 0.8 to match the circles

ldecicco-USGS:
One of the checkboxes in the issue was to "increase opacity on the category legend background rectangle"....this was not trivial for my naive D3 abilities....so, yeah I was under the assumption one of the tasks for this PR was to make it higher than 0.8....

The category hopping issue seems to only be for industrial, which is an improvement from how the jumping was before the "mouse-over" rectangles were created.  I was just creating an issue to see if the "mouse-over" invisible buttons are translated correctly at the end.

jread-usgs:
Sorry, that checkbox wasn't clear. The "legend background rectangle" is the mostly transparent white rectangle that sits behind all of the legend components. I think it is important that the circle color and the legend color be the same if possible, including opacity.

ldecicco-USGS:
OK...in that case...do you have a preference on categoryToColor if the arguments should be category and stroke or category and alpha? We could revert a ton of changes on this pull request

ldecicco-USGS:
Alrighty, I think honestly this is good to go. No transitions, but the code is in there if we want to. The category jumping was being caused by the stroke-width. Adding the `.attr('pointer-events','all')` to those mouse event rectangles seems to fix that. 

desktop

narrow

jread-usgs:
Looks sharp to me!

jread-usgs:
Does `.style('pointer-events','all')` have the same effect?

ldecicco-USGS:
Yup, just double checked on chrome, firefox, ie...things seem to be working across the board

lindsaycarr:
Trying to piece together the conversation here and in #300. Is the final code here fixing that issue or do we still have some more decisions to make around that?

ldecicco-USGS:
The code here is ready to merge. It includes no transitions (although there is commented out code that shows how to start if we want to go down that route). We can merge this and leave 300 open to work on.

PR #277, opened 2018-04-27
added rounded percentages to pie chart

### LINKED ISSUE #262: Spruce up pie chart
- [ ] NWULT: Is it possible to get percentages on the pie chart?  That would be much more informative to the reader, especially with the 5% for “Other” given to the left. JREAD: I think we need to show values for these categories, perhaps as mouseovers or cooked into the callouts. 
- [ ] Could go as far as #159, fancy pie chart animation, but remember the deadline. Maybe just a little movement - expansion from small radius or something like that?

lindsaycarr:
Fixes #262. This adds percentages to text.

image

image

I easily made the slices expand out from the middle, but realized that we would need to get fancier about how to start that animation (scrolling, etc). So, decided against it. Given the time and the other issues, I'm comfortable with this minimal sprucing.

### DIFF
--- a/js/national_pie.js
+++ b/js/national_pie.js
@@ -1,6 +1,6 @@
 // code to add a static pie chart of the national view
 function loadPie() {
-  var width = 480,
+  var width = 525,
       height = 350,
       radius = Math.min(width, height) / 3;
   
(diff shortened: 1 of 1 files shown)


jread-usgs:
lucky it adds up to 100% ;)

lindsaycarr:
@jread-usgs yep - thought of that exact thing

mwernimont:
All about the %

PR #278, opened 2018-04-27
tweak test to see if it's embed or not

### LINKED ISSUE #219: Only run the main figure functions for embed

lindsaycarr:
Fixes #219.

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -29,7 +29,7 @@ var waterUseViz = {
   stateAbrvs: [], // created in extractNames()
   nationalData: {},
   stateData: {},
-  isNotEmbed: !RegExp("html").test(window.location.pathname)
+  isEmbed: RegExp("embed-water-use-15").test(window.location.pathname)
 };
 
 // Globals not yet in waterUseViz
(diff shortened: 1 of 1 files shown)


js/build_map.js
```diff
@@ -29,7 +29,7 @@ var waterUseViz = {
   stateAbrvs: [], // created in extractNames()
   nationalData: {},
   stateData: {},
-  isNotEmbed: (window.location.pathname === "/")
+  isNotEmbed: !RegExp("html").test(window.location.pathname)
```
aappling-usgs: I think this is an improvement but still not foolproof, because http://dev-owi.usgs.gov/vizlab/water-use-15/index.html is a link that produces the page that someone could reasonably choose. What about testing for the full embed page name?
```js
isNotEmbed: !RegExp("embed-water-use-15.html").test(window.location.pathname)
```

I'd also sorta prefer to see a variable called `isEmbed` rather than `isNotEmbed` throughout, but I can live with either.

lindsaycarr: Oh, I guess the index.html one exists, but none of our clickable things gets to that. Yes, I thought about isEmbed, too rather than double negative but would've had to change the tests. I can do that.

aappling-usgs: Yeah, it's a relatively low risk, but seems worth the small change if the test for the full page name "embed-water-use-15.html" works (i hope it will - does `.` being a wild card mess things up with `RegExp()`?)

PR #279, opened 2018-04-27
Remove animations from mobile and testable mobile detection improvement

mwernimont:


js/resize.js
```diff
@@ -8,13 +8,34 @@ function resize() {
...
+     // check for any-pointer:coarse which mostly means touchscreen
+     result = true;
+   } else if (window.TouchEvent || ('ontouchstart' in window)) {
+     // last resort - check for exposed touch events API / event handler
+     result = true;
+   }
+ }
+ 
+ if(result === true){
+   if(screen.width >= 768 && screen.height >= 1024){
```
aappling-usgs: I had worked myself around to 
```
if(Math.max(screen.width, screen.height) > 1024)
```

We need > rather than >= so we don't include assume hovering on ipads, and i think doing it as a max of height & width means we don't need to worry about which orientation it's in.

Are you persuaded by that? happy to hear an argument for keeping what you have if you're confident in it.

js/resize.js
```diff
@@ -8,13 +8,34 @@ function resize() {
...
+     waterUseViz.interactionMode = 'hover';
+   }else{
+     waterUseViz.interactionMode = 'tap';
+   }
+ }else{
+   alert('hover');
```
aappling-usgs: can you remove the alert now?

js/map_utils.js
```diff
@@ -212,8 +212,14 @@ function applyZoomAndStyle(newView) {
     .style("stroke-width",  1/zoom.s); // make all counties have scaled stroke-width
   
   // apply the transform (i.e., actually zoom in or out)
+  var zoomTime;
+  if(waterUseViz.interactionMode === 'hover'){
+    zoomTime = 750;
+  } else {
+    zoomTime = 0;
+  }
```
aappling-usgs: cool, so we'll just have these 3 if-else chunks to edit if we decide to go back to animated transitions on mobile :+1:

mwernimont:
No arguments from me, I didn't think I had it set up that great and was kinda hoping you'd chime in.

Issue #260, opened 2018-04-26
Fiddle with county interaction


aappling-usgs:
- [x] Lighten the county highlight color. county highlight is too subtle

~~- [ ] think about the clicking vs hovering behavior based on these NWULT comments:~~ won't be fixing, see comment below https://github.com/USGS-VIZLAB/water-use-15/issues/260#issuecomment-388145273
  - "On the PC, you can click on a category or the total, then over over counties to see the values on the left.  One logical expectation is that clicking on the county will bring up more data, like the popup that was in an earlier draft.  Instead, this behavior zooms back to the US extent.  (Nancy) I found I did this repeatedly even after noting that clicking a county wasn’t going to do what I expected."
  - "If you do click on the county you are interested in (instead of just observing the county name and data value on the explanation, it zooms to the US extend and is left displaying whatever county your mouse happens to be over in the new extent.  This is pretty confusing too."~~

~~- [ ] In mobile, retain the county selection when you switch categories (irrelevant to desktop): "On the phone version, I tried clicking directly on a county and it would display the Total. Tried then clicking on another category, but the data switched to State total."~~ This is actually just an issue on iOS devices. See #321

mwernimont:
![screen shot 2018-04-27 at 4 44 03 pm](https://user-images.githubusercontent.com/6390096/39386430-387f29ce-4a3a-11e8-9047-162b0401c550.png)
That a good enough contrast?

mwernimont:
I'm also wondering if this a task to add the tooltips back in?  I know it just says think about the comments, but they are basically asking for tooltips on a click.

aappling-usgs:
* That highlighted county looks excessively pale on my screen. What about making it darker than the focal state rather than lighter?

* for item 2, @lindsaycarr and I discussed NWULT's request for tooltips and decided to say no to that - so no, don't add tooltips back in. The goal here is to acknowledge that NWULT found the current behavior (zooming out if you click on a county within the active state) unintuitive, and to think about whether another behavior would be less frustrating. I see three OK options (but there may well be more) for behavior when you click on a county within the active state - (1) status quo = zoom out to national, (2) do nothing at all, which might be less frustrating to someone determined to see more about a county, or (3) change the clicked-on county so that it's permanently active - i.e., hovering stops working and it just keeps showing this active county until you pan or zoom.

[...]

PR #281, opened 2018-04-27
County Highlight change

mwernimont:
Quick PR for the county highlight


lindsaycarr:
Color change for #260

Issue #261, opened 2018-04-26
Text edits


aappling-usgs:
- [x] Figure 1: Explanation should be “U.S. Water Withdrawals” not “U.S. Water Use” ("Explanation" = "legend")
- [x] Figure 1: Explanation “(millions of gallons per day)” should be “(million gallons per day)”
- [x] Figure 2: Insert “withdrawals” into this sentence: “The national breakdown of water withdrawals is as follows:”
- [x] Figure 2: Add "of water" to "The U.S. states and territories used over 322 billion gallons [of water] per day in 2015. "
- [x] Figure 2: Change "over" to "nearly" or "about" in "The U.S. states and territories used [XXoverXX nearly] 322 billion gallons [of water] per day in 2015. "
- [x] Figure 2: “U.S. states” reads “United States states”, so instead maybe “The U.S., including the territories of Puerto Rico and the U.S. Virgin Islands,” 
- [x] Figure 2: “We use water in different ways to serve many economically and culturally important needs.” We all felt “cultural” was not the right word. Perhaps… “We use water in different ways to meet our everyday domestic needs and to serve many important economic purposes.”
- [x] Figure 2: “Together” is misspelled on the pie chart blurb “Domestic, livestock, aquaculture, and mining compose the "other" category and togther make up approximately 5 percent of total withdrawals.”
- [x] Figure 2-3: This sentence is long, consider breaking it up: “While you can still see that irrigation and thermoelectric are the largest, there is a stark contrast between the East, where thermoelectric is dominant, and the West, where irrigation is dominant.” 
- [x] Figure 2-3: “stark contrast” - we see “stark” having negative connotations. Maybe “striking contrast”
- [x] Figure 3: Consider adding a figure caption to the pies map
- [x] Figure 4: Add "blue" to "Drag a [blue] state over its matching bar", and color the word blue w/ the appropriate blue color
- [x] Bottom blurb and references: Remove this sentence: “The USGS Water Availability and Use Science Program hopes to produce higher spatial resolution and more frequent reports in the future”, replace with a separate paragraph or bullet saying “More information on the [USGS Water Availability and Use Science Program](https://www.usgs.gov/water-resources/water-availability-and-use-science-program).”
- [x] Bottom blurb: Remove “accurately” from this sentence: “This information is essential to [XaccuratelyX] understand how future water demands will be met”

aappling-usgs:
[...]

PR #284, opened 2018-04-28
small text change request

lindsaycarr:
Alison requested a small change in #273 to consistency with parentheticals. I edited so both used dashes. This is the rest of #261.

image


lindsaycarr:
Taking liberties and merging this tiny edit myself.

Issue #255, opened 2018-04-26
Medium-res county boundaries for mobile


aappling-usgs:
Alternative to implementing #189 is to use something less than the current high res, more than the current low res, for mobile.

aappling-usgs:
Could be low-res across continental US and medium-res for islands

lindsaycarr:
Hmmm I'm not seeing outlines (even crude ones) for two of the VI islands on mobile or desktop.

![image](https://user-images.githubusercontent.com/13220910/39413238-06671244-4bed-11e8-9e17-eb7b94ba8b5f.png)

aappling-usgs:
The low res version used for desktop at national view does not have those islands at all (the circles exist but not the boundaries), but I thought we were OK with that because they become visible once you zoom in. I am seeing the boundaries for those islands on beta right now - maybe you just need to reload after setting developer tools to mobile?

jread-usgs:
I see them on actual mobile (my phone) on beta, but I do not see them on devtool's version of mobile  even w/ a completely fresh session. Is the mobile detection based on some part of the device that chrome's tool isn't triggering?

jread-usgs:
I think we should consider this done

lindsaycarr:
Lindsay's comment above https://github.com/USGS-VIZLAB/water-use-15/issues/255#issuecomment-385299122 is going to be fixed a la #316.

PR #285, opened 2018-04-29
Medium res

aappling-usgs:
Made medium resolution versions of the state and county boundaries (_mobile.json suffix) that are lots smaller than the high res:
image
image

AND don't give up whole counties in the island territories:
High res:
image
image

New mobile res:
image
image


aappling-usgs:
I think we could do even better on mobile file size with a little more time - in particular by using low-res on the mainland and higher res for the islands - but my plane is landing and rental car will be taking off soon, so this is probably it for the week.

aappling-usgs:
This PR makes good progress toward #255, maybe all the progress we're going to make unless someone else picks up this issue to take it farther.

js/build_map.js
```diff
@@ -90,11 +90,19 @@ var tooltipDiv = d3.select("body").append("div")
   .classed("tooltip hidden", true);
 
 // Read data and add to map
-d3.queue()
-  .defer(d3.json, "data/state_boundaries_USA.json")
-  .defer(d3.tsv, "data/county_centroids_wu.tsv")
-  .defer(d3.json, "data/wu_data_15_range.json")
-  .await(fillMap);
+if(waterUseViz.interactionMode === 'tap') {
```
lindsaycarr: I think there's a cleaner way to show this queue by having shared files not in the if statements. I'm submitting a PR soon, so I'll do that with that PR.

Issue #268, opened 2018-04-27
order of function eval bug


jread-usgs:
I'm not sure I have the expertise to figure this out, and I am also not sure what impact (if any) this has on the viz, but we often get an error on page load:

![image](https://user-images.githubusercontent.com/2349007/39361234-20d3a550-49e7-11e8-8686-b632540fc90e.png)
within `updateLegendTextToView()`
here:
```js
var state_data = waterUseViz.stateData
      .filter(function(d) { 
        return d.abrv === activeView; 
    });
```
(the error is `map_utils.js?_c=7941337:301 Uncaught TypeError: waterUseViz.stateData.filter is not a function`)

ldecicco-USGS:
Gah, that means the async loading still isn't in the right order. I've learned that that error means it still has null for the stateData.  I haven't seen that error since my big pr a couple days ago....but that was the bulk of my frustration for the legend task. I'll keep thinking about it, and maybe one of our D3 experts will realize why it's still loading wrong. Do you happen to know if the error only comes when the browser doesn't have a hash? Or is it just once in awhile?

aappling-usgs:
I thought I had fixed this with the `if()` statements at https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L217 and https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/build_map.js#L229, and I can't reproduce it right now, but I see that those commits happened 5 days ago whereas this issue was created 2 days ago, and I was able to reproduce it on some branch, some build, details now lost, this morning. So I'm not sure whether those changes fully fix the problem.

lindsaycarr:
I'm getting this consistently when the URL I go to is for a state rather than loading the US first and then clicking the state. It seems to be because `addStates` ends up calling updating the view (which eventually calls `updateLegendTextToView` but `waterUseViz.stateData` is loaded later in the `fillMap` function. 

`wu_state_data.json` and `wu_data_15_sum.json` are really small files so I think this can just be fixed by having those part of the initial queue. I'm going to try that.

lindsaycarr:
This was fixed in #286.

PR #286, opened 2018-04-30
move state and national summary json into queue

lindsaycarr:
There are issues with the order in which things are located that has been causing errors, see https://github.com/USGS-VIZLAB/water-use-15/issues/268#issuecomment-385294530. Since these files are small and won't significantly impact the initial load time, I moved them into the overall queue and then rearranged the queue from what is was in #285 so that the `if`-else statement didn't repeat file paths. I no longer see the error mentioned in #268. 

This also required me to add `national_pie` and `rank_states_utils` as resources to the main fig since `loadPie` and `rankEm` are called in `fillMap`. I think we were lucking out that they were inside `d3.json` calls which allowed enough time for those functions to become available when their resources were loaded later down the page.

### DIFF
--- a/viz.yaml
+++ b/viz.yaml
@@ -246,6 +246,8 @@ publish:
       map_states: map_states
       map_counties: "map_counties"
       dropdown_selector: "dropdown_selector"
+      national_pie: "national_pie"
+      rank_states_utils: "rank_states_utils"
       state_boundaries_USA: state_boundaries_USA_topojson
       state_boundaries_zoom: state_boundaries_zoom_topojson
       state_boundaries_mobile: state_boundaries_mobile_topojson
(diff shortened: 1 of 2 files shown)


lindsaycarr:
I'm going to merge this, check out dev, then deploy to beta. I will leave #268 open until others can test it out, too.

viz.yaml
```diff
@@ -246,6 +246,8 @@ publish:
       map_states: map_states
       map_counties: "map_counties"
       dropdown_selector: "dropdown_selector"
+      national_pie: "national_pie"
+      rank_states_utils: "rank_states_utils"
```
aappling-usgs: Should we delete these dependencies from https://github.com/USGS-VIZLAB/water-use-15/pull/286/files#diff-181bf6189cfb48aabd24db45de8426a8R397 now that they're called directly in build_map.js?

lindsaycarr: Oh, good point. I'll get a PR up, but I'm going to sign off for tonight and say that we are in a good place for reviewers right now.

PR #290, opened 2018-04-30
scaling up PR _before_ the simplification

jread-usgs:


js/custom_projection.js
```diff
@@ -101,7 +101,7 @@ function albersUsaTerritories() {
     lower48.scale(_);
     alaska.scale(_ * 0.35);
     hawaii.scale(_);
-    puertoRico.scale(_ * 3);
```
aappling-usgs: ohhh, i get it. cool!

jread-usgs:
related to #285 
adds the VIs to the national view map

PR #294, opened 2018-04-30
welcome back circles-as-paths + fix missing ones

lindsaycarr:
As noted in https://github.com/USGS-VIZLAB/water-use-15/issues/119#issuecomment-385260864, there were some counties that were missing. This was because of using `.select('.wu-path')` instead of `.selectAll('.wu-path')` by mistake. So, it was only adding data for the first half of counties and not the second. This also follows #289 and removes stroke for circles as paths.

image

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -200,10 +200,10 @@ function fillMap() {
   
   // add the circles
   // CIRCLES-AS-CIRCLES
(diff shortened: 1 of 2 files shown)


jread-usgs:
Looks good to merge, but let's keep it off of beta for the time being

PR #295, opened 2018-04-30
fixes #291

### LINKED ISSUE #291: clean up pre-process scaling code that alters PR/VI
Created some tech debt in `process_zip_to_shp.R` that should be cleaned up.

jread-usgs:
This is clean up for #291 and doesn't alter anything now (it isn't used). 

The syntax for using this (PR example):
```yaml
  county_boundaries_PR_shifted:
    command: scale_shifted_shps(county_boundaries_census_sp, fips = I(c('72','78')), scale = I(3))
```

The reason this exists is that we simplify all of the counties and state borders using lat/lon geometries from their actual location, then move some and scale some of them up or down for display (AK and PR notably). So, we are oversimplifying PR relative to its final display (it gets scaled up) and undersimplifying AK relative to its final display (it gets scaled down) when trying to hit the sweet spot with scaling factors so that the 1) file size isn't too big, 2) we don't simplify so much that we drop counties/islands. 

This potential solution may be useful in the future but is probably not now. There is a lot of challenging things going on in the [custom projection](https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/custom_projection.js) js code that I couldn't wrap my head around, but the hope is we can pre-scale and then be able to take the scaling out of js (see [here](https://github.com/USGS-VIZLAB/water-use-15/blob/master/js/custom_projection.js#L104) ) or ...maybe both pre-scale and pre-shift sometime further down the road. 

red is original and black is scaled: 
image

image

I don't think this code fully works with the AK islands that are further west than -180° longitude, but I am not noodling around with that right now because I don't think this is going to be used in the near-term.

### DIFF
--- a/scripts/preprocess/process_zip_to_shp.R
+++ b/scripts/preprocess/process_zip_to_shp.R
@@ -24,15 +24,17 @@ read_shp_zip <- function(zipfile) {
   return(data_out)  
 }
 
-scale_shifted_shps <- function(sp, PR_scale = 3){
+#' @param fips a character vector of fips to be scaled
+#' @param scale a scale factor to apply to fips
+scale_shifted_shps <- function(sp, fips, scale){
   
-  pr_sp <- sp[(sp@data$STATEFP %in% c('72','78')), ]
-  pr_cent <- rgeos::gCentroid(pr_sp, byid=F)@coords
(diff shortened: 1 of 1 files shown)


jread-usgs:
_If_ we are going to end up using this code, we should probably extend this issue to properly handle Alaska.

PR #301, opened 2018-05-01
New rounding

ldecicco-USGS:
Only sigfig the characters, and do both state and national at 6 digits


ldecicco-USGS:
ACK...branch fail

PR #303, opened 2018-05-01
rank'em events 

wdwatkins:
Records dropping off the bars, or right or wrong answers.   No hover events

### DIFF
--- a/js/rank_states_utils.js
+++ b/js/rank_states_utils.js
@@ -175,14 +175,23 @@ function rankEm(barData) {
   rankSvg.droppedState = function(d){
     rankSvg.isDragged = false;
     var barchoice = d3.select('.chosen-rank-bar');
-        
+    var sessionId = getSessionId();
+    var timestamp = getTimestamp();
+    var draggedState = d3.select(this).attr('id').split('-')[0];
     if (barchoice.empty()){
       d3.select(this)
         .transition().duration(600)
           .attr('transform','translate(0,0)'); 
+      
+      gtag('event', 'dropped state off bars' , {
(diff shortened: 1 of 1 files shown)


js/rank_states_utils.js
```diff
@@ -191,7 +200,12 @@ function rankEm(barData) {
...
             .style('opacity', 0);
-        
+        gtag('event', 'dropped state correctly' , {
+          'event_category': 'figure',
+          'event_label': draggedState,
+          'sessionId': sessionId,
+          'timestamp': timestamp
+        });
         // is this the last one? if so, remove the directions    
         if (svgStates.select('#ranked-states-bars').selectAll('rect').filter('*:not(.locked-rank-bar)').empty()){
```
lindsaycarr: wondering if we should stick an event in here - 'completed game' or something?

wdwatkins: We could, although we should be able to determine that based on number of correct state drops.

lindsaycarr: is that easy to summarize in the analytics data that comes back for each user? So it would be simple to know how many users completed the game?

wdwatkins: Hmm, simple in R with a little dplyr, although thinking about it now, it might be good to be able to see that quickly through the GA UI.  I will add it.  I was trying to not get too out of hand with logging events.

lindsaycarr: Yeah, that makes sense. Just wanted to double check that we are making it easy for ourselves to get the information we want! I will merge once the new code pops up here.

PR #317, opened 2018-05-08
National pie chart is nicer

### LINKED ISSUE #312: clean up pie fig
- [ ] not callouts for big ones
- [ ] get "other" categories on chart somehow

also think about how to make it look really nice - this is the second figure people will see. Will it make them stay or leave?

lindsaycarr:
Fixes #312. This pie chart is the second figure people will see and might make the difference between staying or leaving. As such, it deserved some attention and customization.

# Before
image

# After
image


lindsaycarr:
Talked with @jread-usgs and we decided to rearrange so that the "other" categories were easier to read. Result:

image

PR #320, opened 2018-05-08
trying to fix clip extent for PR [probably not merging]

lindsaycarr:
We were missing a tiny island off the west of PR. This code brings it back into view, but then it was overlapping FL, so I had to translate all of PR/VI to the east. Unfortunately, this was no getting cut off by a clip extent there. I tried a combo of widening the clipExtent for lower48 and puertoRico but didn't have any luck. 

I wanted to make a PR so we could return to this sometime else if we wanted to. I will leave it up for a bit and if people have an idea, then we can fix and merge. Otherwise, we aren't going to worry about it.

Here's what this code gets you (VI is slightly cut off):
image

### DIFF
--- a/js/custom_projection.js
+++ b/js/custom_projection.js
@@ -78,7 +78,7 @@ function albersUsaTerritories() {
 
     return (y >= 0.120 && y < 0.234 && x >= -0.425 && x < -0.214 ? alaska
         : y >= 0.166 && y < 0.234 && x >= -0.214 && x < -0.115 ? hawaii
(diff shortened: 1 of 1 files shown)


ldecicco-USGS:
Mona:

image

lindsaycarr:
You've been to our tiny missing island @ldecicco-USGS ?! 😎

ldecicco-USGS:
Yup! The "Galapagos of the Caribbean":
https://islandsofpuertorico.com/isla-de-mona/

lindsaycarr:
Won't fix this time around. The island has a population of 0 and part of a county that  is already shown. We will talk about how to add it to our reusable mapping stuff later.

PR #322, opened 2018-05-10
Scroll tracking

### LINKED ISSUE #146: Analytics - Scroll depth tracking for key stories
We'd like to know how far down a user scrolls before leaving page, with "how far down" pegged to which key story or text section is the last one visible

lindsaycarr:
Fixes #146. I just used existing ids on various divs - I think that is reasonable. This only records an event if you scroll and can see the bottom of the div. I console logged the div id in the gif below to show that in action:

scroll_tagging

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -151,7 +151,7 @@ function customizeCaption() {
       "Hover over the map for details. Click a button to switch categories. " +
       "Click a state to zoom in, and click the same state to zoom out.";
   }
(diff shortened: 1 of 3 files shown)


wdwatkins:
Looks fine to me — do the events only fire the first time you scroll past?  

You can use the 'GA Debug' chrome plugin btw if you want to see the actual events firing.  

What did you use for the screen capture?

PR #326, opened 2018-05-11
Image swapping

mwernimont:
Was scratching my head at why the pie map figure disappeared on mobile thinking I had done something terribly wrong JS wise, and low and behold it was my old CSS.

The CSS can be removed and have deleted the two older mobile images.  The functionality to have the images swap is still being worked on by me, but in the meantime, I don't want people to worry about the figure completely gone on mobile.  So this should fix that issue and I'll have the image swapping functionality by next week Monday COB or sooner. Hopefully sooner, just looking for a solid option, as people offer pretty janky ones on stack overflow.


viz.yaml
```diff
@@ -429,14 +420,11 @@ publish:
...
-    depends: ["styles", "placeholder", "static_story_figure", 
-              "static_story_figure_mobile_left", 
-              "static_story_figure_mobile_right"]
+    depends: ["styles", "placeholder", "static_story_figure", "static_story_figure_mobile"]
     context:
       placeholder: "placeholder"
       pie_map_plot: "static_story_figure"
-      mobile_left_pie_plot: "static_story_figure_mobile_left"
-      mobile_right_pie_plot: "static_story_figure_mobile_right"
+      mobile_pie_map_plot: "static_story_figure_mobile"
```
lindsaycarr: are there mustache template changes that reflect the replacement of `mobile_left_pie_plot` and `mobile_right_pie_plot` with `mobile_pie_map_plot`?

mwernimont: Nope, they were added as a dependent to a mustache file. Those were simply images.

mwernimont:
Bleh, forgot to branch...but found a pretty simple solution to image swapping based on size.  The HTML5 `<picture>` element allows us to set many different sources and put the browser in charge of picking which one it needs based on attributes we tell it to look for.  In this case the media attribute.  It also has a backup of a regular image tag for the 2 browsers that don't support the picture element (IE11 and below, and Opera).

I was skeptical at first that the picture element wasn't just loading all the images anyway with the entire picture element and its sources being in the DOM, but after network tests, it really only does load the image it needs.  That combined with lazy loading it only grabs the image it needs when it needs to show it.  

It's a lot like the favicon solution we created before, and I feel this will work well in the future if not for swapping complete images but having the ability to serve different resolution photos seamlessly.

viz.yaml
mwernimont: Or the change is reflected in my next commit as you were right I did not commit the mustache file for whatever reason...

jread-usgs:
Looks like we need to publish the placeholder gif, right? https://beta-owi.usgs.gov/vizlab/water-use-15/images/placeHolder.gif ?

mwernimont:
It already should be publishing...I have it in the viz.yaml and it ends up in my local target space. Hmm...

mwernimont:
https://github.com/USGS-VIZLAB/water-use-15/blob/master/viz.yaml#L427

It is added as a dependency to the WU_story section and has the context, the image also lives in the repo...

mwernimont:
@aappling-usgs any thoughts on why it wouldn't be able to find it?

aappling-usgs:
Hmm. Locally, I'm also seeing it get moved to target/ and made available to the localhost. Next hypothesis: it's in the new version of vizlab and beta doesn't have that update yet...I'll take a look.

aappling-usgs:
Nope, y'all were on top of the vizlab version update. Next hypothesis: aws s3 sync isn't transferring gifs correctly. Stay tuned...

aappling-usgs:
Well, I tried modifying the sync commands to call out gifs as images specifically:
```
# s3 sync 
source ~/.bash_profile
aws s3 sync ./target/ s3://dev-owi.usgs.gov/vizlab/water-use-15/ --exclude "*.gif" --exclude "*.svg" --exclude "*.json"
aws s3 sync ./target/ s3://dev-owi.usgs.gov/vizlab/water-use-15/ --exclude "*" --include "*.gif" --content-type "image/gif"
aws s3 sync ./target/ s3://dev-owi.usgs.gov/vizlab/water-use-15/ --exclude "*" --include "*.svg" --content-type "image/svg+xml"
aws s3 sync ./target/ s3://dev-owi.usgs.gov/vizlab/water-use-15/ --exclude "*" --include "*.json" --content-type "application/json"
```
but I've reverted this now because I think getting the capitalization right was what was needed (#328). I'm no longer seeing the 404 for placeHolder.gif (or placeholder.gif) on dev or beta.

Issue #53, opened 2018-03-01
Create thumbnails


ldecicco-USGS:

lindsaycarr:
See ImageMagik script used in Hurricane Irma - https://github.com/USGS-VIZLAB/hurricane-irma/issues/110#issuecomment-329227314

aappling-usgs:
We need a first thumbnail or GIF in time for the vizzy submission (~April 13)

aappling-usgs:
Starter code for an R-based GIF is at https://github.com/USGS-VIZLAB/viz-scratch/tree/master/water_use_gifs, looks something like this:

![water_use_2015_animation_pub_irr_pies](https://user-images.githubusercontent.com/12039957/38432475-655bf694-397c-11e8-9013-c8301a51cfb8.gif)

lindsaycarr:
@ldecicco-USGS you are assigned this, but that was a month ago. Still interested or should we reassign to someone else? Also, the code to generate a static national pie map now lives in this repo (see https://github.com/USGS-VIZLAB/water-use-15/blob/master/piemap.yaml), and I think we can reuse it to make various thumbnail sizes.

ldecicco-USGS:
How about we open up the issue to someone for now. I'm going to spend the morning to see if I can get the original vizzy up-to-date.

lindsaycarr:
Tool to check if they are working properly: https://help.medium.com/hc/en-us/articles/215769058-Update-Facebook-Twitter-cards

lindsaycarr:
Facebook looks good, but has two warnings:

![image](https://user-images.githubusercontent.com/13220910/40305772-0035c788-5cca-11e8-8eb5-49ae5ca0b96e.png)

```
Invalid OG URL: The given value '' for property 'og:url' was not a valid absolute URL.
```

Twitter is being cut off slightly at the bottom, so I need to adjust the size:
![image](https://user-images.githubusercontent.com/13220910/40305837-3db0ddf0-5cca-11e8-8869-7c208e0918a0.png)

lindsaycarr:
Twitter actually needs a 2:1 image (see https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image).

lindsaycarr:
Ugh, following the directions for image size, but still cutting if off..... 👊 

[...]

PR #327, opened 2018-05-11
Thumbnails

lindsaycarr:
Part of #53, but I don't want to close that until we verify twitter and facebook cards are working (and we need this out on beta + the actual images on the s3 bucket).

run `remake::make(remake_file = "thumbnails.yaml")` to create these locally yourself


thumbnails.yaml
```diff
@@ -0,0 +1,68 @@
+packages:
```
jread-usgs: is there a reason thumbnails should be in a separate remake file instead of being part of viz.yaml? I would think they would follow the `process`, `visualize`, `publish` chain

viz.yaml
```diff
@@ -273,7 +273,9 @@ publish:
       county_boundaries_mobile: county_boundaries_mobile_topojson
       county_centroids_wu: "county_centroids_wu_tsv"
       wu_data_15_range:  "wu_data_15_range_json"
-      thumb-facebook: thumb-facebook
+      thumb_facebook: thumb_facebook
```
jread-usgs: see comment above about generating these as part of viz.yaml

viz.yaml
```diff
@@ -492,11 +494,27 @@ publish:
...
+    location: images/thumb_facebook.png
+    relpath: images/thumb_facebook.png
     mimetype: image/png
-    thumbType: ["facebook","twitter"]
     title: "Facebook Thumbnail"
     alttext: "Facebook Thumbnail"
+  -
+    id: thumb_twitter
+    publisher: thumbnail
+    location: images/thumb_twitter.png
```
jread-usgs: _if_ you go w/ my suggestion to have these generated as part of the `vizmake()` process, I think the `images/thumb_{}.png` should be gitignored.

thumbnails.yaml
lindsaycarr: Since I was using the code from the national pie map figure generation which is a separate remake, I just kept the same pattern. I can't think of why that would be a problem, so I can do that.

lindsaycarr: well actually, didn't we keep that original map one separate since it's using different spatial data. If that isn't a reason to keep them separate then what is keeping us from adding both the national map figure generation and gif generation to the regular viz yaml? I'm kind of leaning towards not now that I remembered that spatial thing.

PR #329, opened 2018-05-15
Mobile national pie map expanding text functionality

mwernimont:
Have the text and expanding and collapsing functionality for the national pie map mobile text.

A few things to note:

I really didn't want to be loading in a tiny new JS file so did a tricky rename of the dropdown_selector.js to mobile_ui.js to allow this bit of functionality to make sense in the file.

The current functionality has no transition or fanciness.  We may want to consider using jQuery on the next viz if we want easy to implement transitions and effects as they are built in and jQuery generally cuts the amount of code to write JS down by half.  I don't think we should add it to this one as we are pretty bloated as is.

### DIFF
--- a/js/mobile_ui.js
+++ b/js/mobile_ui.js
@@ -208,10 +208,10 @@ window.addEventListener('load', function(event){
   expandInteract.onclick = function(){
     if(contentToShow.style.display === "block"){
       contentToShow.style.display="none";
-      expandInteract.innerHTML = 'Expand to read more'
+      expandInteract.innerHTML = 'Read More'
     }else{
       contentToShow.style.display="block";
-      expandInteract.innerHTML = 'Click to hide'
+      expandInteract.innerHTML = 'Hide'
     }
     
   };
--- a/js/mobile_ui.js
+++ b/js/mobile_ui.js
@@ -200,3 +200,21 @@ function resetCountySelector() {
         return i === 0; 
       });
 }
+
+//Expand National Pie Mapp mobile text
+window.addEventListener('load', function(event){
+  var expandInteract = document.getElementById('national-static-pie-mobile-interact');
+  var contentToShow = document.getElementById('national-static-pie-mobile-content');
+  expandInteract.onclick = function(){
+    if(contentToShow.style.display === "block"){
+      contentToShow.style.display="none";
+      expandInteract.innerHTML = 'Expand to read more'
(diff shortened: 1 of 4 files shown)


js/mobile_ui.js
```diff
@@ -200,3 +200,21 @@ function resetCountySelector() {
...
 }
+
+//Expand National Pie Mapp mobile text
+window.addEventListener('load', function(event){
+  var expandInteract = document.getElementById('national-static-pie-mobile-interact');
+  var contentToShow = document.getElementById('national-static-pie-mobile-content');
+  expandInteract.onclick = function(){
+    if(contentToShow.style.display === "block"){
+      contentToShow.style.display="none";
+      expandInteract.innerHTML = 'Expand to read more'
```
aappling-usgs: What if this just said "Read more" and "Hide" - would that still work?

layout/templates/wu_story.mustache
```diff
@@ -43,7 +43,36 @@
...
+          <p>
+            <strong>6.</strong> The Gulf Coast is another hotspot of <strong>industrial</strong> withdrawals for the chemical and petroleum industries.
+          </p>
+          <p>
+            <strong>7. Thermoelectric</strong> power plants use steam to drive turbines and generate electricity. In the <strong>eastern U.S.</strong> where water is relatively abundant, large volumes of water often are withdrawn, used once for cooling, then returned to the source a little warmer than before. In the <strong>western U.S.</strong>, cooling water is more often withdrawn and recirculated many times, so less is withdrawn overall.
+  
+          </p>
+        </div>
+        
+        <p id="national-static-pie-mobile-interact" class="readMore"><strong>Expand to read more</strong></p>
```
aappling-usgs: Again, I suspect that just "Read more" would work, though I haven't looked at the rendered version.

mwernimont: The language doesn't really affect the functionality, so if you guys want less text it works for me.  For the CO River drought Viz we just had Read More and Hide swapping out and heard no complaints.

aappling-usgs:
I forgot that Lindsay is out of town all week. @mwernimont , could you just shorten the Read More and Hide text and then I'll merge? @lindsaycarr - please add comments or new issues later if needed.

mwernimont:
@aappling-usgs there ya go!

PR #337, opened 2018-05-16
footer template and lazy loading footer images

mwernimont:
Copied over the footer HTML from vizlab and created a `footer.mustache` file.  Luckily we can just add the `lib-footer-css` as a dependency to it and it still goes and gets it.

I had to add the thumbnails as a dependency and give a context to the thumbnails but not the vizlab atom, which seems odd.  Think it's a local cache or something?  I watch it go get it over the network though so it knows where to look for it.

Hope to figure out a cleaner solution for lazy loading footer images in the future, but with the time we have this is an acceptable concession.


jread-usgs:
Cool. Why did the .svgs need to be included?

aappling-usgs:
Submitting my review in the form of a PR to marty's branch. Includes the following changes relevant to this PR:
* github.svg and atom.svg don't need to be copied over from vizlab as long as they're identified in the `depends` and `context` sections of the footer_section vizlab item
* renamed hurricane_marie.png to hurricane_maria.png because poor Marie, and then renamed hurricane_maria item to hurricane_maria_thumb to have the same structure as the water_use_thumb item
* moved footer thumb dependencies from the fullpage item into the footer_section item; the page creation works either way, but it makes more sense to me to have the dependencies be listed under the section where they're actually used

and also these unrelated but minor changes:
* removing the `inline: true` line from the selectArrows item is enough to eliminate all 3 warning messages about how we use svginject now and inline is deprecated
* quotes around "other" weren't rendering for me in the new mobile explanation text, so I changed the curly quotes into html curly quote codes

Issue #339, opened 2018-05-17
refine white space


aappling-usgs:
* consider making our text narrower throughout (i.e., prototype the "skinny text proposal")
* reduce the white space above and below the big pie

aappling-usgs:
I've experimented with making our text narrower. I think it's not a good idea now that I've tried it, but I'll push a PR that allows us to change our mind about that later anyway.

Here's what's on beta right now (before any white space manipulation):
![image](https://user-images.githubusercontent.com/12039957/40209126-75a012b0-59f2-11e8-915e-b3240f7e7f45.png)

Here's the status quo after reducing white space above and below the big pie:
![image](https://user-images.githubusercontent.com/12039957/40209057-1c3819e8-59f2-11e8-972a-cd96caa10151.png)

And here's the effect of reducing text width to 600px:
![image](https://user-images.githubusercontent.com/12039957/40209030-fa035586-59f1-11e8-9874-67c106c17626.png)

I thought it might look good, but from this experiment I think the skinny text doesn't work well, maybe because our figures already have a lot of white space and non-rectangular arrangement.

PR #340, opened 2018-05-18
White space

aappling-usgs:
Working on, and possibly resolving, #339 refine white space. @mwernimont , this is new territory to me - please check my css carefully for noob mistakes.

* reduced white space around the big pie
* experimented with making the text skinnier than the figures, but reverted (see #339 for figures explaining why)
* centered the two figure captions in mobile/skinny mode

Remaining/new issues:
* Locally I'm seeing different number rounding, compared to beta, in the county map legend. For example, the national total is xxx,xxx.0 rather than being an integer. Did I do this with my css changes?? I don't think so because I just rebuild from the master branch and am getting the same thing. Maybe it'll be different on dev.
* Internet explorer. It looks OK but has more white space around the pie map than Chrome does. @mwernimont any ideas? chrome on the left, IE on the right:
image

### DIFF
--- a/layout/templates/map_section.mustache
+++ b/layout/templates/map_section.mustache
@@ -1,8 +1,8 @@
 <div class="main-svg">
 </div>
 
-<div id="fig-caption" class="viz-text">
-  <p>[Caption goes here]</p>
+<div id="map-caption" class="caption full-width">
+  <p></p>
 </div>
 
 {{{ map_script }}}
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -152,7 +152,7 @@ function customizeCaption() {
       "Click a state to zoom in, and click the same state to zoom out.";
   }
   
-  d3.select('#fig-caption p')
+  d3.select('#map-caption p')
     .text(captionText);
 }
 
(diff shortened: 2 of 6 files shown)


jread-usgs:
New whit space looks good! I ran into that issue with seeing different rounding when I made the national pie gif. Not sure where the change came from but it is in that national sum json file.

jread-usgs:
but now I am not seeing the same rounding issue...back to whole numbers.

aappling-usgs:
I wonder if an old/new version of that national sum json file has been floating around - could explain why i had to add https://github.com/USGS-VIZLAB/water-use-15/pull/340/commits/77bed2ee00881d81becbaa09dd1f25a39d88d02f for this PR even though it's apparently been working for others already.

layout/templates/map_section.mustache
```diff
@@ -1,8 +1,8 @@
 <div class="main-svg">
 </div>
 
-<div id="fig-caption" class="viz-text">
-  <p>[Caption goes here]</p>
```
jread-usgs: thank you

PR #341, opened 2018-05-21
Adjust twitter card image size + url warning from Facebook

lindsaycarr:
More for #53. Still need to merge, and get the changes to the images that are on the prod URL to test before closing the issue. 

Twitter was getting cutoff (see https://github.com/USGS-VIZLAB/water-use-15/issues/53#issuecomment-390633732)


lindsaycarr:
This will need to wait and merge once my national pie fig PR is merged so I can recreate the twitter thumbnail here to use the correct data.

lindsaycarr:
OK, this is good to review then merge when it's approved.

PR #342, opened 2018-05-21
Outdated data fix

### LINKED ISSUE #330: is national pie map using old data?
I noticed that `piemap` and `thumbnails` use cache/county_centroids_wu.json, which I think is an older version of the file but doesn't exist anymore if you have cleared your cache recently. 
I think we now have two files that need to be combined to contain that info: county_centroids_USA.json and county_centroids_wu.tsv	

This may not be an issue _unless_ this _wu.json file predates the xls update of the county data, in which case it would differ from the newer file(s) in some ways.

lindsaycarr:
Fixes #330. Can't believe I missed this but thanks @jread-usgs for catching! This needs to get merged before #341 so that I can update the new Twitter image there. Plus @mwernimont will need to use the new national fig map.

I followed https://github.com/USGS-VIZLAB/water-use-15/blob/master/scripts/gifs/data_utils.R#L2-L35 to update this.


jread-usgs:
Doe the national map image not end up checked into the repo?

lindsaycarr:
No, it ends up in the `figures/` folder which isn't git tracked

PR #343, opened 2018-05-21
remove hovers for mobile devices

### LINKED ISSUE #321: strange click behavior on apple phones
1) on mobile national view tap a non-total category, 2) tap a state and expect to zoom into it, 3) instead see that the category has switched back to total. 

This happens on @jread-usgs phone (iPhone), but not on @lindsaycarr phone (Samsung). Must have something to do with how iOS handles clicks/mouseover/mouseouts?

lindsaycarr:
Will fix #321, but don't want to close that issue until it's tested. I'm not sure the best way to test this other than to get it out on prod so that someone with an iPhone can verify the behavior is fixed. Problem was that we kept `mouseover` events active even on mobile devices and thought it wouldn't matter, but iOS apparently registers hovers. Decided to remove those and I think that will fix the weird behavior.

### DIFF
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -41,10 +41,14 @@ function addButtons() {
       updateCategory(d.toLowerCase(), activeCategory);
     })
     .on('mouseover', function(d){
-      showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
+      if(waterUseViz.interactionMode === "tap") {
+        showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
+      }
     })
     .on('mouseout', function(d){
-      showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
+      if(waterUseViz.interactionMode === "tap") {
+        showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
+      }
     });
   
   // button category labels
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -32,24 +32,24 @@ function addButtons() {
     .attr('width', waterUseViz.dims.buttonBox.width);
   
   // button rectangles for *mouse events*
-  buttons.append('rect')
+  var mouser = buttons.append('rect')
     .classed('mouser-button', true)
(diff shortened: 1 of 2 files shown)


js/buttons.js
```diff
@@ -41,10 +41,14 @@ function addButtons() {
...
+      if(waterUseViz.interactionMode === "tap") {
+        showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
+      }
     })
     .on('mouseout', function(d){
-      showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
+      if(waterUseViz.interactionMode === "tap") {
+        showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
+      }
     });
```
aappling-usgs: 1. The conditionals you've added make it so there's only hover action if `(waterUseViz.interactionMode === "tap")`...but isn't 'tap' equivalent to 'mobile', where hovering isn't really a thing? Shouldn't you be testing for `==="hover"` instead?
2. The chaining syntax is nice, but this looks to me like a place where breaking the syntax might be cleaner so that you only set any sort of 'mouseover' or 'mouseout' event if hovering is an option. Like this:
```js
  var mouser = buttons.append('rect')
     .classed('mouser-button', true)
     .style('opacity','0')
     .style('stroke-width',1.5)
     .style('pointer-events','all')
     .on('click', function(d){
        updateCategory(d.toLowerCase(), activeCategory);
      });

  if(waterUseViz.interactionMode === "hover") {
    mouser
      .on('mouseover', function(d){
        showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
      })
      .on('mouseout', function(d){
        showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
      });
  }
```

lindsaycarr: Oof nice catch on # 1 - really thought about that too and still picked the wrong one. Fixes coming for # 2

lindsaycarr:
OK @aappling-usgs, I think I've got it now :)

PR #344, opened 2018-05-21
New images and remove .DS_Store

mwernimont:
Pretty straightforward with the images

wu_stories_static_mobile
wu_stories_static

I updated the `.gitignore` to stop watching `.DS_Store` files, which is a MAC exclusive annoyance.


jread-usgs:
Looks fine to me. Is it better to use jpeg over png for these types of images?

mwernimont:
So PNG's are larger file sizes due to the transparency ability they possess and the lossless compression, since we have no transparency and we want smaller files sizes I would say jpg is the way to go.

Issue #346, opened 2018-05-22
remove mouseout from category selector?


jread-usgs:
I think the "preview" category mouseover may result in confusion. Instead, mousing over a category could be the same as clicking the category (it sets the active category).

aappling-usgs:
Wow, I was completely blind to this possibility for confusion before, but makes sense now that I've played with it on dev with this proposal in mind - I agree that it'd be more intuitive for the category to stay on whichever one you hovered over most recently.

jread-usgs:
And does removing the mouse out event also help us resolve the iOS tap issue?

lindsaycarr:
@jread-usgs I don't think it would help with that directly, but we're going to implement this as part of the solution for that.

lindsaycarr:
@jread-usgs should be fixed, but we didn't get a chance to push to beta to check. Probably can do that on 5/25 since Mindi no longer needs beta to be stable.

aappling-usgs:
Should be on beta now - @jread-usgs , could you check it out on your iOS?

jread-usgs:
The category selector is fixed, but tapping a county does not highlight it.

aappling-usgs:
Hmm, I can reproduce that on my phone and on my desktop with mobile developer tools, too. I think I can fix this - PR coming soon.

PR #347, opened 2018-05-22
Hovers broke from #343

lindsaycarr:
updateCategory hasn't actually been doing anything to change the categories...it's all been the hover. 

lost hovers because when `waterUseViz.interactionMode` was checked in the if statement for addButtons, it was undefined.

### DIFF
--- a/js/map_utils.js
+++ b/js/map_utils.js
@@ -240,7 +240,7 @@ function updateCategory(category, prevCategory) {
   
   // update page info
   setHash('category', category);
-  documentCategorySwitch(category, prevCategory, action = "click");
+  showCategory(category, prevCategory, action = "click");
 }
 
 function showCategory(category, prevCategory, action) {
--- a/js/buttons.js
+++ b/js/buttons.js
@@ -32,24 +32,26 @@ function addButtons() {
     .attr('width', waterUseViz.dims.buttonBox.width);
   
   // button rectangles for *mouse events*
-  var mouser = buttons.append('rect')
+  buttons.append('rect')
     .classed('mouser-button', true)
     .style('opacity','0')
     .style('stroke-width',1.5)
     .style('pointer-events','all')
     .on('click', function(d){
       updateCategory(d.toLowerCase(), activeCategory);
-    });
-  
-  if(waterUseViz.interactionMode === "hover") {
-    mouser
-      .on('mouseover', function(d){
+    })
+    .on('mouseover', function(d){
+      if(waterUseViz.interactionMode === "hover") { 
+        // need these ifs inside .on("mouseover") because waterUseViz.interactionMode 
+        // is undefined when addButtons is run
         showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
-      })
-      .on('mouseout', function(d){
+      }
+    })
+    .on('mouseout', function(d){
+      if(waterUseViz.interactionMode === "hover") {
         showCategory(activeCategory, d.toLowerCase(), action = 'mouseout');
-      });
-  }
+      }
+    });
   
   // button category labels
   buttons.append('text')


js/map_utils.js
```diff
@@ -240,7 +240,7 @@ function updateCategory(category, prevCategory) {
   
   // update page info
   setHash('category', category);
-  documentCategorySwitch(category, prevCategory, action = "click");
+  showCategory(category, prevCategory, action = "click");
```
lindsaycarr: not losing the event documentation because `showCategory` calls `documentCategorySwitch`

aappling-usgs:
There's a lot going on here, and I'm not sure I understand fully, but I think I want to propose a very different approach to this.

The problem, as I currently (barely) understand it, is that iOS mobile devices recognize a hover event and let that interfere with a click event when you tap with your finger. On desktop, we're relying on hover + click to complete the steps needed to fully change categories. So executing only those commands triggered by hover works poorly. I don't know what's happening on non-iOS mobile devices and am worried that this means I'm missing something important. Still, forging ahead:

The solution proposed in this PR is to
1. reintroduce a hover event on all mobile devices, even though most of them can't recognize hover and the ones that do recognize it do the wrong thing with it, and
2. cause updateCategory to call showCategory such that on desktop, showCategory will get called at least twice when you move the mouse onto a button and then click.

Neither of these steps seem ideal to me, for the reasons noted above.

Instead, how about:
1. separate into its own function the code (currently in resize.js) that detects waterUseViz.viewport and waterUseViz.interactionMode. call this, say, `detectDevice`.
2. call `detectDevice` early in build_map.js, such that the value of `waterUseViz.interactionMode` is defined before `addButtons` gets called.
3. don't define mouseover or mouseout events unless `waterUseViz.interactionMode === "hover"` (i.e., revert to #343)
4. maintain the separation between updateCategory and showCategory that existed before this PR.
5. if `waterUseViz.interactionMode === "hover"`, define `mouser.on('click'` as it's currently being defined, like this:
    ```
    .on('click', function(d){
      updateCategory(d.toLowerCase(), activeCategory);
    });
    ```
6. if `waterUseViz.interactionMode === "tap"`, define `mouser.on('click'` as it's currently being defined, like this:
    ```
    .on('click', function(d){
      showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
      updateCategory(d.toLowerCase(), activeCategory);
    });
    ```

lindsaycarr:
Here's a more concise summary of the behavior I'm seeing. On beta: iOS registers hovers when it shouldn't (and on the mobile view in dev tools, clicking a category also registers as a hover); desktop technically registers a hover before a click anyway. After adding `if` so mouseover events are only added when interactionMode == "hover" (result of #343), showCategory was no longer called on mobile, so only the hash changed when you clicked and nothing was visually updated.

Potential solutions are complicated by #346. I don't really want to get into change waterUseViz.interactionMode earlier, but I guess if it wasn't too complex we could do that, or we could leave the `ifs` inside the actions.

Potential solution 1 (change the URL hash when button is clicked, only visually update when mobile click):
```
.on('click', function(d){
  updateCategory(d.toLowerCase(), activeCategory);
  if(waterUseViz.interactionMode === "tap") { 
	showCategory(d.toLowerCase(), activeCategory, action = 'click');
  }
})
```

Potential solution 2 (if we decide #346 is a good idea, don't register click at all for desktop and have mouseovers update the hash and change the active category):
```
.on('click', function(d){
  if(waterUseViz.interactionMode === "tap") { 
    updateCategory(d.toLowerCase(), activeCategory);
    showCategory(d.toLowerCase(), activeCategory, action = 'click');
  }
})
.on('mouseover', function(d){
  if(waterUseViz.interactionMode === "hover") { 
    // need these ifs inside .on("mouseover") because waterUseViz.interactionMode 
    // is undefined when addButtons is run
    updateCategory(d.toLowerCase(), activeCategory);
    showCategory(d.toLowerCase(), activeCategory, action = 'mouseover');
  }
});
// and delete mouseout action
```

For either of these, I would want to add the ability to turn off events for showCategory or updateCategory because they both fire an event right now.

js/map_utils.js
```diff
@@ -231,27 +231,25 @@ function applyZoomAndStyle(newView, doTransition) {
...
-    updateButtons(category);
-    updateButtonWidths(category);
-    updateCircleCategory(category);
-    if(action !== "mouseout") {
-      documentCategorySwitch(category, prevCategory, action);
-    }
-  }
+  
+  // change buttons and map circles
+  updateButtons(category);
```
aappling-usgs: The one surprise to me here is the removal of `if(prevCategory !== category`)`, which I think would still have utility here so that we're not updating buttons and circles any more than strictly necessary. Am I missing something that makes this check obsolete now?

lindsaycarr: updateCategory has an if statement right at the top that makes this redundant

```
if(category === prevCategory) {
    // don't do anything if you click/hover on current category
     return;
}
```

aappling-usgs: oh, so it does! thanks.

Issue #335, opened 2018-05-16
use d3 modules rather than whole d3


aappling-usgs:
d3 is by far the largest file - can we use just some components?

aappling-usgs:
first step: determine which modules we need.
1) can we test to make sure that those are exactly and completely the list of needed modules?
2) can we test to see whether it actually saves us time, and by how much? (probably only makes things faster if we can also be using webpack, #334)

jread-usgs:
Not digging into this yet, but the resource I found when I started looking earlier in the week was this: https://stackoverflow.com/questions/44379393/how-can-i-specify-the-d3-v4-micro-library-on-cdn

jread-usgs:
I took a quick stab at stitching the microlibraries we are using into a single .js. It seems to work until we hit `queue`, which seems different than the other modules and also seems to be something that has gone away in v5 of d3 (maybe why its microlibrary is different). From my initial look it seems that we can cut the .js file size down by ~50%, which seems like it _might_ be worth the extra effort, but I am not sure. We could cut that down even more by only exporting the functions from each microlibrary we are using but that is more work too. 
The approach that seems to be the path forward looks like this: https://towardsdatascience.com/d3-js-the-custom-modular-bundle-now-bebd6f25bc8b and https://bl.ocks.org/mbostock/bb09af4c39c79cffcde4

aappling-usgs:
Do you have a list of the microlibraries that you found necessary (recognizing that `queue` isn't quite captured yet)?

jread-usgs:
This is likely incomplete, but:
d3-array, d3-color, d3-collection, d3-scale, d3-geo, d3-selection, and then [d3-queue](https://d3js.org/d3-queue.v1.js) is the one I had trouble with (that is the link to the non-minified). The other microlibraries export things (e.g., [d3-selection](https://d3js.org/d3-selection.v1.js) ), but queue doesn't and I end up with something like `build_map.js?_c=4391381:94 Uncaught TypeError: d3.queue is not a function`

jread-usgs:
I had a conversation with Mary about this, and they are using broserify and it sounds like webpack is likely a better solution for us, but actually much more complicated and has a greater learning curve.

mbucknell:
It's browserify

jread-usgs:
hahah. bro. whoops

jread-usgs:
and seems the issue w/ queue might be related to big changes from v4->v5 of d3, where promises are the new norm.

[...]

PR #350, opened 2018-05-23
Don't use d3.queue, just nest

lindsaycarr:
Setting up to modularize the d3 libraries + load a map view before circles load to speed up "time to first paint". Part of #335.

### DIFF
--- a/js/build_map.js
+++ b/js/build_map.js
@@ -91,18 +91,29 @@ var tooltipDiv = d3.select("body").append("div")
   .classed("tooltip hidden", true);
 
 // Read data and add to map
-var dataQueue = d3.queue();
+var stateDataFile;
 if(waterUseViz.interactionMode === 'tap') {
-  dataQueue.defer(d3.json, "data/state_boundaries_mobile.json");
+  stateDataFile = "data/state_boundaries_mobile.json";
 } else {
-  dataQueue.defer(d3.json, "data/state_boundaries_USA.json");
+  stateDataFile = "data/state_boundaries_USA.json";
 }
-dataQueue
-  .defer(d3.tsv, "data/county_centroids_wu.tsv")
-  .defer(d3.json, "data/wu_data_15_range.json")
-  .defer(d3.json, "data/wu_data_15_sum.json")
-  .defer(d3.json, "data/wu_state_data.json")
-  .await(fillMap);
+
+d3.json(stateDataFile, function(error, data1) {
+	if (error) throw error;
+  d3.tsv("data/county_centroids_wu.tsv", function(error, data2) {
+	  if (error) throw error;
+    d3.json("data/wu_data_15_range.json", function(error, data3) {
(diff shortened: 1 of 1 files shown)


js/build_map.js
```diff
@@ -91,18 +91,29 @@ var tooltipDiv = d3.select("body").append("div")
...
+      d3.json("data/wu_data_15_sum.json", function(error, data4) {
+	      if (error) throw error;
+        d3.json("data/wu_state_data.json", function(error, data5) {
+	        if (error) throw error;
+          fillMap(data1, data2, data3, data4, data5);
+        });
+      });
+    });
+  });
+});
```
aappling-usgs: very pretty code nest! =)

aappling-usgs: i'm wondering if it would be more readable to assign the read-in data to globals just after these if(error) lines rather than using data1, data2, etc. as placeholders. that way a higher fraction of the variables would have meaningful names, and you wouldn't have to keep track of the order of data file loading within fillMap.

lindsaycarr: so I renamed to something meaningful, but you're saying read as globals? Like put them into waterUseViz?

aappling-usgs: Yeah, I meant put them straight into the globals - see below. But I also don't mean to be a huge pain about this. If you're ready to move on to other tasks, I'm OK with this as in 07609c0.

js/build_map.js
```diff
@@ -98,17 +98,17 @@ if(waterUseViz.interactionMode === 'tap') {
...
 	if (error) throw error;
-  d3.tsv("data/county_centroids_wu.tsv", function(error, data2) {
+  d3.tsv("data/county_centroids_wu.tsv", function(error, countyCentroids) {
 	  if (error) throw error;
-    d3.json("data/wu_data_15_range.json", function(error, data3) {
+    d3.json("data/wu_data_15_range.json", function(error, waterUseRange) {
 	    if (error) throw error;
-      d3.json("data/wu_data_15_sum.json", function(error, data4) {
+      d3.json("data/wu_data_15_sum.json", function(error, waterUseNational) {
 	      if (error) throw error;
```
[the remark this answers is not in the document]
lindsaycarr: OH but are you also suggesting creating globals for the ones we don't put into waterUseViz already? That's where I am a little concerned. I don't want to push stateBoundsUSA into the global thing because it might get complicated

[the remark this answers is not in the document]
lindsaycarr: The way I handle the state bounds and county centroids will change slightly when I work on getting fillMap split into a fxn for making a map and a separate fxn for adding the circles. That will come in a different PR.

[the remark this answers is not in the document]
aappling-usgs: Didn't notice that there were some that were never global - agreed that they should just get passed to `fillMap` as before. I like the change you just made, assigning immediately to globals if and only if they need to be globals.

Issue #157, opened 2018-04-09
508 full compliance


aappling-usgs:
Follows on #152 - we need full compliance for the official vizzy release.

lindsaycarr:
No more errors 👍

PR #351, opened 2018-05-24
Five o eight

aappling-usgs:
There were 4 508 errors and maybe 10 alerts. I got rid of all of the errors, some of which turned into alerts instead, and got rid of one additional alert.

The errors were about:
* form items (the drop-downs for selecting states and counties) needed labels or title attributes. I added title attributes.
* hrefs need either text or, if an image, alt text in the image. We had an href around a div around an image with alt text, so I moved the div to the outside so we had a div around an href around an image with alt text. this satisfied the plug-in.

The alert I fixed was about:
* we were skipping h2 and going straight to h3 and h4. i bumped up the header levels to h2 and h3, respectively, and then altered the css so the font sizes would be ~the same.

I don't think there'll be any visible changes due to the error corrections, but it does appear that I've accidentally added a little white space after "How do we use water in the U.S.?", and it's possible that there will be noticeable font differences on some browsers. I've checked Chrome and IE11 and Firefox and think the whitespace result is still OK.

One change I don't like is that IE is no longer picking up the font on the state rank game's instructions:
image
The font looks fine on Chrome and Firefox, and I've tried some things but am not going to succeed in fixing the IE font without some more help or exploration.

### DIFF
--- a/layout/css/custom.css
+++ b/layout/css/custom.css
@@ -118,8 +118,12 @@
   padding-bottom: 10px;
 }
 
-#wu-story h2{
-  margin-top:20px;
+h2{
+  font-size:1.17em;
+}
+
+h3{
+  font-size:1em;
 }
 
 .story-one,
--- a/layout/css/custom.css
+++ b/layout/css/custom.css
@@ -149,7 +149,7 @@ h3{
   font-size:0.9em;
   font-style: italic;
   text-align: center;
-  color: rgb(137,137,137);
+  color: rgb(106,106,106);
   line-height:1.5em;
   margin-top:20px;
   text-decoration:none;
--- a/layout/css/custom.css
+++ b/layout/css/custom.css
@@ -201,9 +201,9 @@ h3{
   pointer-events:none;
 }
 
-.extras{
+#extras{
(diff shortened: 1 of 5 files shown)


aappling-usgs:
addresses #157, could be enough to resolve it.

aappling-usgs:
Hmm, that's good that you don't see it. If I have a CSS change locally, it's hiding pretty well...I've just deleted my target dir and rebuilt, and that didn't help. And I did Ctrl-F5, which I believe is a hard refresh for IE11, and that didn't help.

lindsaycarr:
Huh ... I just did that, too. I still see the font. Strange - let's have someone else try this.

aappling-usgs:
Ohhh, right, I forgot about the contrast tab. Will check that out after getting jenkins working again.

aappling-usgs:
@lindsaycarr - needs https://github.com/USGS-VIZLAB/vizlab/pull/379, but the last couple of commits should fix the contrast errors, which i think will make us actually done with #157.

aappling-usgs:
Updated Jenkins, so merging now

PR #356, opened 2018-05-25
County selection in mobile

aappling-usgs:
Somehow we lost the ability to tap on counties and see them get highlighted in mobile. This PR restores that capacity.

180525-county-tap


aappling-usgs:
Resolves https://github.com/USGS-VIZLAB/water-use-15/issues/346#issuecomment-392066385

Issue #358, opened 2018-05-30
test out uglify for minifying js


lindsaycarr:
Webpack still in the works, but we are going to try to simple paste all of the js files into one big one, then run uglify (see https://www.rdocumentation.org/packages/js/versions/1.1/topics/uglify) and see if there are improvements to performance.

merge in separate branch and deploy on Jenkins to test performance

lindsaycarr:
Test complete, and we are accepting. Went from about 4.7 seconds to 2.2/2.3 seconds for time to first paint. As jread mentioned in the Slack conversation:

> wow, so 2 seconds of load time saved because we reduced the number of files being loaded...
> yeah, and makes you wonder about the load cost of those 6 *tiny* svgs we are pulling in as external files vs embedding in the DOM (e.g., https://beta-owi.usgs.gov/vizlab/water-use-15/images/github.svg and others)

As a future task, we could look into having the tiny svgs be part of the page template, so that they are part of the DOM and not being loaded in as files.

On Alison's treo before (using East region networks):

![image](https://user-images.githubusercontent.com/13220910/40939185-9018723e-6809-11e8-8b73-5c84d1f7dfa0.png)

On David's treo after (using West region networks):

![image](https://user-images.githubusercontent.com/13220910/40939224-a95b212e-6809-11e8-9a31-e95568076d49.png)

Since we used different networks, David set it up using East network, too. It was very similar:

![image](https://user-images.githubusercontent.com/13220910/40939243-baf447bc-6809-11e8-90b3-9120c3b87631.png)

PR #360, opened 2018-05-31
Uglify test

lindsaycarr:
Still need to push out to beta and do some performance tests. This creates a publish function that combines all of the js files into one and then minifies it using `js::uglify_optimize`. Combining means we go from loading 18 individual files to 2 (`all.min.js` and `build_map.js`). Minifying brings the combined file from 98 KB to 54 KB. 

The `build_map.js` script still has to be standalone because it can't be loaded until after divs have been created (I took out some of the functions and put them in a secondary js that was bundled up called `build_map_utils.js`). This is why you see the script tag farther down the page.

Part of #358.


lindsaycarr:
Ooops - forgot to do a PR to a separate branch.

Issue #334, opened 2018-05-16
bundle and minify js files


aappling-usgs:
potentially use two bundles: one that's critical, one that's non-critical. add async to the non-critical one

MVP could be to create a single bundle, postprocess using webpack or similar

PR #361, opened 2018-05-31
Uglify test (to new branch)

lindsaycarr:
Still need to push out to beta and do some performance tests. This creates a publish function that combines all of the js files into one and then minifies it using `js::uglify_optimize`. Combining means we go from loading 18 individual files to 2 (`all.min.js` and `build_map.js`). Minifying brings the combined file from 98 KB to 54 KB. 

The `build_map.js` script still has to be standalone because it can't be loaded until after divs have been created (I took out some of the functions and put them in a secondary js that was bundled up called `build_map_utils.js`). This is why you see the script tag farther down the page.

Part of #358 and also related to #334.


jread-usgs:
I see a 20~25% improvement in initial load time w/ this (emulating a 3G connection locally). :+1:

jread-usgs:
Note we also have 6 CSS files that are also taking up multiple connections and likely should be rolled up into one as well.

viz.yaml
```diff
@@ -294,74 +280,22 @@ publish:
...
+      js_files: [
+        "js/url_hashes.js", "js/custom_projection.js", "js/styles.js", "js/buttons.js",
+        "js/resize.js", "js/map_utils.js", "js/states.js", "js/counties.js", "js/watermark.js", 
+        "js/circles.js", "js/mobile_ui.js", "js/modernizr_custom.js", "js/lazy_loading.js", 
+        "js/national_pie.js", "js/rank_states_utils.js", "js/scroll_tracking.js", "js/build_map_utils.js"
+      ]
+    publisher: combine_minify_js
+    scripts: [scripts/publish/combine_minify_js.R]
   -
     id: build_map
```
jread-usgs: I realize this is a small script and I understand the reason why this file is separate from the others, but should `build_map` be minified as well?

lindsaycarr: Oh yeah, I can do that!

jread-usgs:
Interesting that all of the custom js (minus build map) is larger than the state borders topojson: 
image

Neat that the borders are that small.

lindsaycarr:
We really have 6? Huh wow, we do. I'll do some quick research into that separately.

jread-usgs:
yes, but note the 6 includes some (one or two?) that are coming out of the vizlab package.

lindsaycarr:
actually, 4 are from vizlab. This viz only has `custom.css` and `map.css`

lindsaycarr:
OK, merging. Any CSS bundling will come separately.

PR #362, opened 2018-06-01
Bundle and minify CSS

### LINKED ISSUE #336: bundle and minify css files
is this the same solution as #334? if so, resolve both at once; otherwise, prioritize just getting something working for #334 and see if we can learn any more about the potential in future efforts

lindsaycarr:
Fixes #336.

Bundled them up into one file (goes from 13.5KB for 6 files to 12KB for 1), but does not minify. I tried minifying manually using requireJS, but it only removed 1KB (I followed http://requirejs.org/docs/optimization.html#onecss) so I'm not going to worry about getting it hooked up.

Another side effect was that the embed fig wasn't getting the correct CSS and now it is since they are all bundled.

Currently:
image

This PR:
image

### DIFF
--- a/scripts/publish/combine_minify_css.R
+++ b/scripts/publish/combine_minify_css.R
@@ -1,0 +1,42 @@
+publish.combine_minify_css <- function(viz) {
+  args <- viz[["publish_args"]]
+  
+  # make sure to eliminate this one if it exists already
+  if(file.exists(viz[["location"]])) { 
+    file.remove(viz[["location"]])
+  }
+  
+  # combine
+  file_connection <- file(viz[["location"]])
+  
+  # first add vizlab resource css
+  for(lib in args[["lib_css"]]) {
+    lib_viz <- as.viz(lib)
+    publish(lib)
+    lib_fp <- paste0(exportLocation(), lib_viz[["location"]])
+    css_f <- readLines(lib_fp)
+    write(css_f, viz[["location"]], append = TRUE)
+    file.remove(lib_fp)
+  }
+  
+  # then add cutom css for this viz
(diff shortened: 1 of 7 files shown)


scripts/publish/combine_minify_css.R
```diff
@@ -0,0 +1,44 @@
+publish.combine_minify_css <- function(viz) {
```
jread-usgs: sorry. getting picky about names. But if this doesn't minify, it should be `combine_css`

aappling-usgs:
So the embedded version now has margins? Seems like a lot to get picky about, but if we're serious about encouraging WSCs to use those embedded versions, we probably don't want the margins.

lindsaycarr:
@aappling-usgs I think it's because of what we are choosing to embed. Proabably shouldn't be the whole map div, but should just be the svg. Let me see if I can quick fix it

scripts/publish/combine_minify_css.R
lindsaycarr: true, I did leave the code that we would use to minify in here commented out in case we want to reference it in the future. That's why I didn't change the name. But I see your point - let me know if you still think that's a good idea

jread-usgs: I'll defer to you. I don't want to hold up a PR because of a name change.

lindsaycarr:
Ok, so the div class `.main-svg` is actually a part of vizlab, so I just changed it around so that the embed div does not get that class.

Now embed looks like this:
image

scripts/publish/combine_minify_css.R
lindsaycarr: easy fix, so I'll change it

PR #363, opened 2018-06-01
add tweet and FB prompts

lindsaycarr:
Fixes #52.

image

capture

^ links and image cards don't show up because the site is not active

PR #365, opened 2018-06-04
Merge uglify-test into master

### LINKED ISSUE #334: bundle and minify js files

lindsaycarr:
This fixes #334 and #336


lindsaycarr:
This branch was updated with master up until #357, so the only thing I'm concerned about is #363 but those changes to the social media URLs aren't included in the additions/subtractions for my changes, so I think it will be fine.

lindsaycarr:
I was able to pull down master, build locally, and start up the viz on a local server and it looks/functions as I expect 👍

PR #369, opened 2018-06-08
add topojson and d3 js files to all.min.js

### LINKED ISSUE #366: combine all js files into one
No reason not to combine the d3 js file(s), and files that go into `all.min.js` into one big master minified js file.

lindsaycarr:
Fixes #366. Combined topojson (6.4 KB), d3.v4.min (214.9 KB), and all.min (54.1 KB) resulting minified all.min.js is 275.3 KB. So the gain here is that it removes two more files to load and frees up slots for other things that need to load.

Curious to look and see if this affects performance at all!

### DIFF
--- a/scripts/publish/combine_minify_js.R
+++ b/scripts/publish/combine_minify_js.R
@@ -8,6 +8,20 @@ publish.combine_minify_js <- function(viz) {
   
   # combine
   file_connection <- file(viz[["location"]])
+  
+  # first add vizlab resources
+  for(lib in args[["lib_js"]]) {
+    lib_viz <- as.viz(lib)
+    publish(lib)
+    # vizlab location puts these files in js/d3-modules/
+    # so need to change just to js/
+    lib_fp <- paste0(exportLocation(), "js/", basename(lib_viz[["location"]]))
+    js_f <- readLines(lib_fp)
+    write(js_f, viz[["location"]], append = TRUE)
+    file.remove(lib_fp)
+  }
+  
+  # then add custom js
   for(fp in args[["js_files"]]) {
     js_f <- readLines(fp)
     write(js_f, viz[["location"]], append = TRUE)
--- a/viz.yaml
+++ b/viz.yaml
@@ -261,8 +261,8 @@ publish:
       thumb_landing: thumb_landing
       wu_state_data_json: wu_state_data_json
     context:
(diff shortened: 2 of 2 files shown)


scripts/publish/combine_minify_js.R
```diff
@@ -8,6 +8,20 @@ publish.combine_minify_js <- function(viz) {
   
   # combine
   file_connection <- file(viz[["location"]])
+  
+  # first add vizlab resources
```
jread-usgs: I think this is fine for this viz, but in the future, we probably want it to loop through js resources in the order they were declared, as opposed to these being first because they are libraries. Here, we want them first anyways, but this kind of changes the order behind the scenes as opposed to giving the js library order control to the args in the viz.yaml

scripts/publish/combine_minify_js.R
```diff
@@ -8,6 +8,20 @@ publish.combine_minify_js <- function(viz) {
...
+  for(lib in args[["lib_js"]]) {
+    lib_viz <- as.viz(lib)
+    publish(lib)
+    # vizlab location puts these files in js/d3-modules/
+    # so need to change just to js/
+    lib_fp <- paste0(exportLocation(), "js/", basename(lib_viz[["location"]]))
```
jread-usgs: no need to change this, but is it better to use `paste0()` over `file.path()` for these situations?

viz.yaml
```diff
@@ -441,7 +435,7 @@ publish:
       wu_data_15_sum_json: wu_data_15_sum_json
       wu_state_data_json: wu_state_data_json
     context:
-      resources: ["lib_d3_js", "topojson", "all_css", "all_js_minified"]
+      resources: ["all_css", "all_js_minified"]
```
jread-usgs: not sure if this matters, but we now have css in front of js for the resources. Is that intentional?

scripts/publish/combine_minify_js.R
lindsaycarr: I couldn't use file path because `exportLocation()` returns a directory with the `/`

lindsaycarr: I guess I could strip that off and then use file.path

viz.yaml
lindsaycarr: oops, not intentional and I think it would be smart to switch them.

PR #383, opened 2018-06-19
Huc based version (merged into a completely new branch that should remain it's own branch)

lindsaycarr:
Not perfect. There are still issues with zoom and with all the paths being drawn. Pretty close, pretty quickly though! Not sure how much farther to take this.

huc_wu_animation


jread-usgs:
This looks good - is there a separate set of code to get the huc json built correctly, or does Toby's format work right out of the box?

lindsaycarr:
Nope - no separate code. I just am using it straight from the Google Drive folder he stored it on

lindsaycarr:
Merging so that the others can use this branch.

PR #385, opened 2018-11-15
get static pie figure back

lindsaycarr:
We lost this around the time we made the blackout thumbnails with the blue solid circles. We need to make a higher resolution one, so I updated the `piemap.yaml` code to do so. Increasing the lwd seems to make windows render of this slightly more crisp, but I still think macs can do better.

national_pies

### DIFF
--- a/scripts/piemap/draw_utils.R
+++ b/scripts/piemap/draw_utils.R
@@ -8,7 +8,7 @@ make_arc <- function(x0, y0, r, from_angle, to_angle){
 }
 
 
-plot_national_pies <- function(us_states, us_dots, metadata, filename, watermark_file = NULL){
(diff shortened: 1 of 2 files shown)


scripts/piemap/draw_utils.R
```diff
@@ -69,6 +89,7 @@ plot_slice <- function(x,y,r,angle_from, angle_to, cat, col = NULL){
   polygon(c(x, segments$x, x), c(y, segments$y, y), 
           border = NA,
           col = fill_col(col))
+  lines(segments$x, segments$y, lwd=0.4, col = col)
```
jread-usgs: am assuming this lwd is good for the high-res plot?

PR #388, opened 2019-04-05
PR + USVI gifs

lindsayplatt:
I hacked my way through getting PR to work, then made changes for getting USVI to work and broke PR.... can't spend any more time to get PR working, but I think the PR dots are just not lining up with the polygons. Based on what I've experienced so far with this, the culprit likely lives somewhere in `shift_centroids`. Also, could not get the right authentication to push up to our Google Drive folder of gifs. I manually added them to the folder.

Made a separate makefile for each since they didn't fit the same process as the other states.

```
scmake('gifs/PR_water_use_twitter.gif.ind', remake_file = 'pr_gif.yml')
scmake('gifs/VI_water_use_twitter.gif.ind', remake_file = 'usvi_gif.yml')
``` 

VI_water_use_twitter

PR_water_use_twitter


wdwatkins:
Maybe some package has changed since we last built this that is affecting the style?

lindsayplatt:
It might be because I started with the national gif yaml and edited it. There may have been some style thing I missed.

